PR #5, opened 2015-02-25
Setup: Initial Site Styling + Views  (UX-1600)

talbs:
This work adds in initial rendering/styling support for the Pattern Library site (PL site):

* initial index + individual element styling
* PL site scss setup and abstraction
* tweaks to templates, posts, supporting files
* light JavaScript support for PL site


clrux:
Firstly, what are you look for with review here?

Secondly, a few notes on this particular branch:
* `gulp` no longer works, but `jekyll serve` does.
* However the styling looks incomplete; perhaps something didn't get built/compiled with the lack of `gulp`

Finally, a few questions in general:
* Should we ban/outlaw/murdah the `h5` and `h6`? They're so infrequently used and often smaller than our smallest body copy. I know we could restyle them and all, but, they're just so unused...
* The left-side navigation... some links are in-page (scrolling) and others take you somewhere else. We should clearly identify they are different so folks don't get surprises. This could be different treatments, an icon, or something else.

talbs:
@clrux, this is ready for a review now. Can you please review and confirm the following:

* You can stand up the pattern library locally by following the [updated installation instructions](https://github.com/edx/ux-pattern-library/blob/talbs/pl-styling-init/README.md)
* The rules I've added here are scoped enough to the pattern library site and will not interfere with actual pattern styling we want to port over
* The Markup/DOM structure I've added here match our HTML guidelines - https://github.com/edx/ux-pattern-library/wiki/Styleguide:-HTML
* The rules I've added here match our Sass/CSS guidelines - https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Sass-&-CSS
* The directory structure I've expanded upon here matches our general guidelines - https://github.com/edx/ux-pattern-library/wiki/Styleguide:-General
* You know how and where to create a new element

Some things that will be coming shortly that are not part of this work:

* real elements
* responsive touches to the pattern library layout itself
* navigation improvements (per your feedback and WRT to active/focused states for items)

- - -

FYI, @frrrances

clrux:
@talbs 

**Installation instructions**
* It wasn't clear to me that I needed to re-run `npm install` to update the dependencies. The way this step is described, it's as if it's something one does on first install. Consider updating the text to say something like "update your dependencies every time, just in case"...

**HTML Styleguide**
* Are we including the note about older IE's and browser-specific scoping? If we are targeting IE9+, is this relevant?
* For attribute order, perhaps it makes sense to have `class` and `id` come first, but the others I'm not so picky about and I feel like requiring them in that order would make tests more finicky. (I've had experiences with failed tests because an `id` came before a `class` and it was extremely frustrating). What do you think about only requiring the `class` and `id` order, but leaving the rest in the bag?

** CSS Styleguide**
* You may want to clarify "meaningful use of whitespace;"
* Space between "display" and "related" (currently "displayrelated")
* Consider putting the opening brace in code (i.e., `{`) for readability here
* Consider explaining "@content" and maybe placing it within code (i.e., `@content`)
* Todo's remaining; punting?
* Regarding variable naming, I've been thinking that things like `$blue-l2` isn't very meaningful, and a developer would need to spend time testing the color to see what it was. Will there be some visuals to accompany these variables (perhaps we include little "swatches" on our Pattern Library front-end)? If not, we should consider using more semantic variable names.
* Reconsider the need for older IE's here too.

**Creating new elements**
* It isn't clear how to create new elements here. I'm assuming this means adding to the Pattern Library and if so, it assumes knowledge with Jekyll. How does one create a new "post" for example, and include that post within the navigation?

talbs:
@clrux, thanks for the feedback and apologies for the confusion in what I needed the review on. Here's some feedback on your feedback :P...

* **Updated the README.md** to include a note about ``npm install`` not just being a one time thing. Thanks!
* **Attribute Order** - Agree that class/id are the most important attrs, but I'd prefer to keep the other detailed ordering in for reference. If tests are failing based on attr order, that sounds like a crappy test that should be fixed.
* **IE8 + Older Technique** - I'm keeping these in as this is the preferred way to handle older IE UI issues in the case that it comes up. I've previously made a note that we don't need these conditional comments and won't be using them, but if an open source user needs them (for their user base), I want to recommend the cleanest way to do so.
* **Fixed the typos and code syntax** cases you brought up in the wiki guides
* **Variables + Colors** - yes, it would be good to get some swatches in at some point for values. This isn't needed now.
* **TODOs** - I'd like to keep these in to keep us honest. We can complete/remove them once the work has been done.
* **Creating New Elements** - I'm not going to document how to create new elements in the README.md file now since its just you, me, and our team working on this in the near future. I've provided example elements for reference and copying/pasting. I'm assuming that you and anyone else has [read up on Jekyll](http://jekyllrb.com/docs/home/) and [how to post](http://jekyllrb.com/docs/posts/). If you are having trouble after that, we can chat in person to sort out any wrinkles. I'll add some reference links to the README.md with places to learn up on Jekyll.

clrux:
:+1:

talbs:
@frrrances, I merged this to make sure it wasn't going to block any of this sprint's stories. Feel free to review if you'd still like to. I'll be following up with some extracurricular tweaks anyhow.

PR #7, opened 2015-03-06
UX-1571 Define base theme settings

clrux:
- - -
_Note: this branch is not to be merged. It serves as an example for theming and will be updated over time to reflect additions to the base pattern library partials._
- - -

# Theming the edX platform
The files included in this branch serve as a reference for _how_ to create themes. Creating a theme does the following:
* provides override values to existing variables in the edX platform
* provides additional values for use, extending the platform

## Using the partials
This branch includes a `_utilities.scss` partial and a `_variables.scss` partial. These two files include example styles that override the platform styles. Specifically in `_variables.scss` we're using different brand colors (in this case, specific to MIT), a different typeface, and a few other styles that are to differ from the platform styles.

Only variable or style overrides need to be included in theme partials. Not specifying a value will let theme use the platform styles.

### DIFF
--- a/_src/static/sass/utilities/_variables.scss
+++ b/_src/static/sass/utilities/_variables.scss
@@ -25,12 +25,12 @@ $baseline-horizontal:    $baseline;
 // color: config
 $transparent:       rgba(0, 0, 0, 0);
 
+$blue:              rgb(0, 162, 227);
+$pink:              rgb(183, 37, 103);
 $gray:              rgb(127, 127, 127);
 $red:               rgb(178, 6, 16);
 $yellow:            rgb(237, 189, 60);
 $green:             rgb(37, 184, 90);
-$blue:              rgb(0, 162, 227);
-$pink:              rgb(183, 37, 103);
 
 $brand-primary:     $blue;
 $brand-secondary:   $pink;
--- a/_src/themes/mit/static/sass/_variables.scss
+++ b/_src/themes/mit/static/sass/_variables.scss
@@ -1,0 +1,47 @@
+// ----------------------------
+// MIT (Example file)
+// This file provides utility overrides for base/_utilities in the main platform compile.
+// It allows instutions to provide alternative values when displaying the application.
+// ----------------------------
+
+
+// ----------------------------
+// #COLORS
+// ----------------------------
+$mit-maroon:                    rgba(163, 31, 52, 1.0);
+$dark-gray:                     rgba(138, 139, 140, 1.0);
+$light-gray:                    rgba(194, 192, 191, 1.0);
+
+$brand-primary:                 $mit-maroon;
+$brand-secondary:               $dark-gray;
+$brand-accent:                  $light-gray;
+
+
+// ----------------------------
+// #TYPOGRAPHY
+// ----------------------------
+@import url("http://fonts.googleapis.com/css?family=Lato:400,900,400italic");
+
+// Font families
+$font-family-sans-serif: 'Lato', sans-serif;
+
+
+// ----------------------------
+// #APPLICATIONS
+// ----------------------------
+// Scaffolding
+$view-bg: rgba(235, 235, 235, 1.0);
+
+// Base text settings
+$text-base-color: $dark-gray;
+$text-base-focus-color: $brand-primary;
+
+// Links
+$link-color: $brand-primary;
+$link-focus-color: saturate($brand-primary, 20%);
+$link-visited-color: desaturate($brand-primary, 20%);
+
+// Headings
+$headings-base-color: $dark-gray;
+$headings-primary-color: $black;
+$headings-secondary-color: $light-gray;
--- a/_src/themes/mit/static/sass/_utilities.scss
+++ b/_src/themes/mit/static/sass/_utilities.scss
@@ -5,7 +5,58 @@
 // ----------------------------
 
 
-// reset of links
+// ----------------------------
+// #COLORS
+// ----------------------------
+$mit-maroon:                    rgba(163, 31, 52, 1.0);
+$dark-gray:                     rgba(138, 139, 140, 1.0);
+$light-gray:                    rgba(194, 192, 191, 1.0);
+
+$brand-primary:                 $mit-maroon;
+$brand-secondary:               $dark-gray;
+$brand-accent:                  $light-gray;
+
+
+// ----------------------------
+// #TYPOGRAPHY
+// Font families
+// ----------------------------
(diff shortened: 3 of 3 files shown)


_src/themes/mit/static/sass/_variables.scss
```diff
@@ -0,0 +1,7 @@
+$red:                           rgba(163, 31, 52, 1.0);
+$dark-gray:                     rgba(138, 139, 140, 1.0);
+$light-gray:                    rgba(194, 192, 191, 1.0);
```
clrux: Institutions may supply their own color names and values...

_src/themes/mit/static/sass/_variables.scss
```diff
@@ -0,0 +1,7 @@
...
+
+$brand-primary:                 $red;
```
clrux: ...and then redefine the primary, secondary, and tertiary brand colors which are used throughout the application.

_src/themes/mit/static/sass/_variables.scss
```diff
@@ -0,0 +1,7 @@
...
+$brand-secondary:               $dark-gray;
+$brand-accent:                  $light-gray;
```
clrux: ...and then redefine the primary, secondary, and tertiary brand colors which are used throughout the application.

_src/static/sass/utilities/_variables.scss
```diff
@@ -25,6 +25,8 @@ $baseline-horizontal:    $baseline;
 // color: config
 $transparent:       rgba(0, 0, 0, 0);
 
+$white:             rgb(255, 255, 255);
```
talbs: Hmm, looks like this is already defined in https://github.com/edx/ux-pattern-library/blob/clrux/base-theme-settings/_src/static/sass/utilities/_variables.scss#L98. If the Sass Map makes it hard to customize this value, I'd recommend using the ``palette``function to define a ``$white`` var. Perhaps something like:

```scss
$white: palette(grayscale, white-t);
```

_src/static/sass/utilities/_variables.scss
```diff
@@ -25,6 +25,8 @@ $baseline-horizontal:    $baseline;
...
+$black:             rgb(0, 0, 0);
```
talbs: Hmm, looks like this is already defined in https://github.com/edx/ux-pattern-library/blob/clrux/base-theme-settings/_src%2Fstatic%2Fsass%2Futilities%2F_variables.scss#L104. If the Sass Map makes it hard to customize this value, I'd recommend using the ``palette``function to define a ``$black`` var. Perhaps something like:

```scss
$black: palette(grayscale, black-t);
```

_src/static/sass/utilities/_variables.scss
```diff
@@ -25,6 +25,8 @@ $baseline-horizontal:    $baseline;
...
 $gray:              rgb(127, 127, 127);
```
talbs: Hmm, looks like this is already defined in https://github.com/edx/ux-pattern-library/blob/clrux/base-theme-settings/_src%2Fstatic%2Fsass%2Futilities%2F_variables.scss#L92. If the Sass Map makes it hard to customize this value, I'd recommend using the ``palette``function to define a ``$gray`` var. Perhaps something like:

```scss
$gray: palette(grayscale, base);
```

_src/themes/mit/static/sass/_variables.scss
```diff
@@ -0,0 +1,77 @@
...
+
+
+// ----------------------------
+// #TYPOGRAPHY
+// ----------------------------
+@include font-face('Open Sans', '../fonts/OpenSans/OpenSans-Light-webfont', 300, $file-formats: eot woff ttf svg);
```
talbs: NIT: do we want to try to include a different typeface here to see the cascade working? If so, you could just import from Google Web Fonts to keep it simple and not add other asset files.

_src/static/sass/utilities/_variables.scss
```diff
@@ -25,6 +25,8 @@ $baseline-horizontal:    $baseline;
...
 $red:               rgb(178, 6, 16);
```
talbs: See my notes above about using values already in the Sass color maps. There is one for yellow.

_src/static/sass/utilities/_variables.scss
```diff
@@ -25,6 +25,8 @@ $baseline-horizontal:    $baseline;
...
 $yellow:            rgb(237, 189, 60);
```
talbs: See my notes above about using values already in the Sass color maps. There is one for yellow.

talbs:
@clrux, I made some notes about the variables you added in our base variables.scss partial. Once you address those and add the background/maintenance note to the description of this PR, things look good.

I noticed one color value I need to fix up, I'll do that in a separate PR.

talbs:
:+1: once feedback is addressed.

_src/static/sass/utilities/_variables.scss
```diff
@@ -25,17 +25,9 @@ $baseline-horizontal:    $baseline;
 // color: config
 $transparent:       rgba(0, 0, 0, 0);
 
-$gray:              rgb(127, 127, 127);
```
talbs: These values are needed for the Sass map below. Why did you remove them?

talbs:
Thanks, @clrux for the fixes. Let's squash this into reasonable commits to show your theming abstracting work. My hope is that as we build more patterns, we can add more theming customizations in future commits to this branch to show how they can be overriden (especially until we move things + figure our configuration).

_src/themes/mit/static/sass/_variables.scss
```diff
@@ -0,0 +1,47 @@
+// ----------------------------
```
talbs: @clrux, sorry for the late-breaking comment, but one thing I just noticed is that this file/directory structure doesn't match the one we planned for in https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Sass-&-CSS#main-style-compile. Can you sync up where this is (meaning move it to the ``_utilities.scss`` file of the MIT theme)?

clrux:
@frrrances I've moved the redefined variables into utilities to match what we've spec'ed in our compile.
(cc @talbs)

talbs:
@clrux, this PR's branch has been open for a long time and the theming work here may be at odds with how edX plans to theme its platform down the road. Should this be closed and we re-create a theming example with all of those current events in mind?

clrux:
Absolutely.

clrux:
Closing this outdated example pull request in lieu of a new example using the updated library and features.

PR #10, opened 2015-03-10
UX-1679 Defining heading styles

clrux:
WIP...

Questions:
* Do we want to handle media queries inline like this, or do we want to have a separate file, perhaps for individual resolutions (i.e., _1600.scss, _1024.scss, _768.scss, etc.)?

### DIFF
--- a/_src/static/sass/components/_headings.scss
+++ b/_src/static/sass/components/_headings.scss
@@ -1,0 +1,138 @@
+// ------------------------------
+// edX Pattern Library: Components - Headings
+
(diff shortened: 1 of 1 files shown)


_src/static/sass/base/_reset.scss
```diff
@@ -0,0 +1,82 @@
...
+* {
+    box-sizing: border-box; // Reset the box model for easier management
+}
+
+
+// ----------------------------
+// Base
+// ----------------------------
+body {
+    font-size: 100%;
```
clrux: Set this to 100% rather than a defined value (previously $font-size).

PR #6, opened 2015-03-05
UX-1525 Define base elements

clrux:
_This is a work in progress!_

@talbs I'm to a point where I'd like you to take a look through this. A few things to note:
* I've stripped out most of the styles we were previously using. I want a blank slate with only the basics. I figured we'd continue to add to these as we flesh out the library.
* I changed some of the pre-existing mixins, variables, and extends to have more semantic and clear names. As we or other developers are building views, I wanted the Sass names to be intuitive.
* I'm still pondering the functions. I'd like to get creative with it and maybe have some that handle fonts, colors, and buttons based on supplied variables, but I'm still reading up on it. If you have info, holla.

Thanks!

PR #11, opened 2015-03-11
Talbs: Define Base Elements

talbs:
This work is an extension to https://github.com/edx/ux-pattern-library/pull/6. It offers an alternative base elements set up.


_src/static/sass/utilities/_helpers.scss
```diff
@@ -0,0 +1,139 @@
...
+// About: Sass partial for defining extends and re-usable elements at the Sass compile level.
+
+// #UTILITIES
+// #RESETS
+
+// ----------------------------
+// #UTILITIES
+// ----------------------------
+// visual link
+%fake-link {
```
clrux: What would be the use case for this? Makes me kinda nervous...

_src/static/sass/utilities/_helpers.scss
```diff
@@ -0,0 +1,139 @@
...
+    &:last-child {
+        margin-bottom: 0;
+        padding-bottom: 0;
+        border-bottom: none;
+    }
+}
+
+// accessibility
+%a11y-hide {
+  border: 0;
```
clrux: NIT: Spacing

_src/static/sass/utilities/_helpers.scss
```diff
@@ -0,0 +1,139 @@
...
+    text-indent: 100%;
+    white-space: nowrap;
+}
+
+// ----------------------------
+// #UTILITIES
+// ----------------------------
+// reset of lists
+%no-list {
+  list-style: none;
```
clrux: NIT: Spacing

_src/static/sass/utilities/_helpers.scss
```diff
@@ -0,0 +1,139 @@
...
+    overflow: hidden;
+    text-indent: 100%;
+    white-space: nowrap;
+}
+
+// ----------------------------
+// #UTILITIES
+// ----------------------------
+// reset of lists
+%no-list {
```
clrux: Is `no-list` accurate? This would go _on_ a list, no? I feel like this lacks semantics.

_src/static/sass/utilities/_helpers.scss
```diff
@@ -0,0 +1,139 @@
...
+// reset of code
+%reset-code {
+    font-family: $font-family-monospace;
+    font-size: font-size(base);
+    line-height: line-height(large);
+}
+
+// reset of links
+%link {
+    @include transition(all 0.15s ease-in-out);
```
clrux: Should we use one of the timing variables here?

_src/static/sass/utilities/_variables.scss
```diff
@@ -3,156 +3,316 @@
...
+$pink:              hsla(16%, 50%, 43%, 1.0);
+
+$brand-primary:     $blue;
+$brand-secondary:   $pink;
+$brand-accent:      $gray;
+
+// colors: palettes
+$palettes: (
+    primary:        (
+        base:       $brand-primary,
```
clrux: For each of these colors, how many variations will we really need?

_src/static/sass/utilities/_variables.scss
```diff
@@ -3,156 +3,316 @@
...
+@include font-face('Open Sans', '../fonts/OpenSans/OpenSans-BoldItalic-webfont', 700, italic, $file-formats: eot woff ttf svg);
+
+// font families
+$font-family-sans-serif: 'Open Sans','Helvetica Neue', Helvetica, Arial, sans-serif;
+$font-family-monospace: 'Bitstream Vera Sans Mono', Consolas, Courier, monospace;
+
+// scale roots
+$font-scale-small: 14;
+$font-scale-medium: 16;
+$font-scale-large: 16;
```
clrux: Large is the same as medium; do we need them both?

_src/static/sass/utilities/_variables.scss
```diff
@@ -3,156 +3,316 @@
...
 
-$link:            $blue;
-$link-hover:      $blue-lighter;
-$link-visited:    $blue-dark;
 
+// ----------------------------
+// #TIMING
+// ----------------------------
+$timing: (
+    base:       0.75s,
```
clrux: Do we need this many variations? Maybe for visual consistency we limit it to three: slow, normal, fast?

_src/static/sass/utilities/_helpers.scss
talbs: The main use case is if we need to style something that doesn't natively give us the cursor property like this - Its more to cover our butts for never interactions (think sliders or other elements). Happy to remove it if it really gives you the creeps - right now its not used and this will never make it to our compiled CSS as a Sass extend.

_src/static/sass/utilities/_helpers.scss
talbs: Good feedback - I'll roll the properties here into ``%reset-list`` extend.

_src/static/sass/utilities/_helpers.scss
clrux: I'd prefer to remove it - it's only a single style, after all - and define `cursor: pointer` on an as-needed basis. You cool with that?

_src/static/sass/utilities/_helpers.scss
talbs: Yep, totes!

_src/static/sass/utilities/_variables.scss
talbs: yes, these are roots that will be used for different breakpoint/response font bases - the scale will flex depending on this initial value. So, medium will be used for medium-based viewport sizes and large will be used for large.

_src/static/sass/utilities/_variables.scss
talbs: Good question - I'd prefer us to use them some and see how many we end up referencing. I know in the past, the slower ones have been used for larger UI animations/transitions while the very fast ones are used for basic transitions. Let's revisit as we get further in PL work.

_src/static/sass/utilities/_variables.scss
talbs: Good question - This is creating monochromatic scales of values per color (the actual color base is  defined/abstracted in good old Sass variables beforehand, for easier configuration and theming). I'm fine with pairing down the computed scale once we start using them, but this gives us different shades of each base color (primary, secondary, accent, black-white scale, and system colors like red, green, yellow) which seems barebones enough.

talbs:
@clrux, I've addressed your comments (thanks for the thoughts/feedback). For the items that were straightforward, I've made changes. For the larger questions, I've given my thoughts. Let me know if you have any more recommendations or hesitations on those.

_src/static/sass/utilities/_variables.scss
clrux: What about slimming it down now and seeing if we need more and if so, add them? I fear that if we left them, we'd forget to go back and slim down. Adding as we need ensures that we never had too much, and means we won't have to search/find through the entire app to see if this or that is being used.

talbs: Sure, will do.

_src/static/sass/utilities/_helpers.scss
talbs: Sure, will do.

talbs:
@clrux, I've fixed those two additional requests you had. Mind giving this a look/thumbs for merging?

talbs:
Got the verbal thumbs, so merging. Thanks!

PR #13, opened 2015-03-16
Adding function for color contrast

clrux:
@talbs I've added a couple of functions that I hope will help us generate better contrast with our colors choices. We'd just need to call this function when setting colors for elements and buttons. Untested as of yet, but wanted to get this down before I forgot.

I also adjusted spacing/tabs for what existed previously :-)

Updates:
* Added props to authors
* Cleaned up comments/nots
* Removed two three functions that aren't useful, leaving just the useful ones


clrux:
Closing as I've got another idea.

PR #14, opened 2015-03-18
adding in RTL setting needed by Neat + compiled asset cleanup

talbs:


talbs:
@clrux, I recently did this for the certificate template work in the LMS and thought to add it here. In short, we need to make sure Neat's grid layout is set to RTL in addition to any bi-app margin/padding we add to elements. This work adds that setting (http://thoughtbot.github.io/neat-docs/latest/#default-layout-direction).

It also cleans up some bad code in our compiled CSS - looks like there was a merged conflict with the checked-in compiled files that was not caught during your base elements work. We should make sure we freshly compile those before any commit/merge.

talbs:
@clrux, mind giving this your quick blessing?

clrux:
:+1:

PR #22, opened 2015-03-30
Adding in missing Font Files for Pattern CSS

talbs:


talbs:
@clrux, mind a quick look-see on this fix? I noticed that we had some missing assets loaded into the DOM - found the culprit to be Open Sans for our patterns CSS.

clrux:
:+1:

PR #24, opened 2015-04-01
Updating example text

clrux:
@talbs FYI


clrux:
Closed in lieu of 25.

PR #28, opened 2015-04-02
Fix - Colors Clean Up + Sync

talbs:
This work involves some quick syncing and correcting of color values based on our standards and conventions.


clrux:
:+1: Squash and lgtm.

PR #25, opened 2015-04-01
Elements: Headings

clrux:
This cleaned up branch replaces https://github.com/edx/ux-pattern-library/pull/24

@talbs FYI

PR #29, opened 2015-04-02
Fix - Missing Heading

talbs:
adding in missing heading lvl 5 (and adjusting remainder of scale)


talbs:
@clrux, in doing some visual static designing, I realized one level of headings that was visually designed was missed during https://github.com/edx/ux-pattern-library/pull/25. This is the level 5 heading of really 7 heading styles. 

This work adds it back. Mind reviewing?

clrux:
:+1:

PR #32, opened 2015-04-03
Publish: Tabs, stateful states

clrux:
@talbs This brings the tabs work into `gh-pages`, along with your stateful statefullness.


talbs:
This is a long line of commits. Can you review that all of them are part of the work we want to bring in (or make sense to bring in since the last publish PR)?

talbs:
I think the last publish PR that was merged was https://github.com/edx/ux-pattern-library/pull/26, right?

clrux:
I think all of these are pertinent. They either deal with the colors/grayscale or the tabs. Then there's that CNAME one. That sound/look right?

talbs:
@clrux, I thought we agreed to make a ``publish-*`` branch off of master to make a PR like this against gh-pages. Can you close this and do that, please?

PR #38, opened 2015-04-07
Color pattern revisions and cleanup

clrux:
@talbs This small bit of work removes the duplicate gray values with the merlot saturations.


talbs:
@clrux, the preview site looks good. Thanks for the quick clean up.

Are there Sass variables/key values that need to be removed as well for our base gray (I'm thinking saturated values that aren't needed)?

Let me know on that one detail and I'll give thumbs.

clrux:
@talbs :+1: ?

_src/static/sass/components/_colors.scss
```diff
@@ -232,18 +232,6 @@
         background: palette(accent, x-dark);
     }
 
-    &.mid-sat {
```
talbs: These are being removed from brand-accent. I believe we still want to have the saturated values for this configured color, no? Remember, a company/org may not have gray as their accent like we do.

clrux: In our particular PL case, the grays that had the saturation was our brand accent. So, maybe we need to rename that gray to be neutral or something?

talbs:
@clrux, here's a quick PR into your branch to revise some content/markup - https://github.com/edx/ux-pattern-library/pull/39. 

Feel free to merge once you review and have no issues. Also, you can fixup both of these commits into yours when you're ready.

talbs:
@clrux, another quick PR for a few colors that you planned for in the palette rendering but never made it into the Jekyll post - https://github.com/edx/ux-pattern-library/pull/40. I also slightly tweaked the black + white values.

As you and the rest of the team mentioned, we'll need a color audit shortly :P

talbs:
@clrux, please:

1. rename this PR to something that covers what we're doing here (color pattern revisions/cleanup) better 
2. squash any of my "fixup!"-prefixed commits with yours 
3. adjust your commit message to describe what was done by us here (the current one is out of date)

After all that is done, :+1:

PR #41, opened 2015-04-08
JS: checking for palettes that don't provide enough contrast

clrux:
@talbs I created this script which places a yellow highlight around those swatch values that fail to meet or exceed AA contrast specifications. This should allow us to easily weed out or avoid using those color combinations.

Caveats:
* The script only checks for contrast for body copy. Many of these combinations _may_ work when the text is of a larger size or bolded.
* For those values that have an alpha channel, the alpha channel is discarded and the color tile is flattened. In most cases these palettes pass because the background is faded back enough to make the text readable.
* We may be able to modify this script for use in sandboxes, highlighting all elements that fail. It might make it easier for us to see visually what we need to address.


talbs:
@clrux, I'm taking a look now, but HEADS UP: the commits you originally had looked funny, so I pulled this branch and locally rebased against a fresh copy of master. **I've force-pushed to this remote/PR and the commits look much better**.

Since I force-pushed, please pull locally before you do any more work or try to push yourself.

talbs:
:thumbs once you squash.

PR #42, opened 2015-04-08
PL: Revising Styling and Infrastructure for (color) Warning UI

talbs:


talbs:
@clrux, here are a few tweaks to the warning UI work you started in https://github.com/edx/ux-pattern-library/pull/41. I know why you placed the ``!important`` rules next to the warning properties, but I'd prefer not to get in that game of using those if you don't mind. 

If you feel strongly, we can add them back in as a warning's style should override anything else if its present. 

Let me know what you think.

clrux:
@talbs Were you able to get the font color to be black without the `!important`? I found that it was receiving the background color, but the font color was still the base font color. Not a deal-breaker, just not what I was going for.

talbs:
@clrux, no, not without the ``!important`` added. Since our actual pattern CSS trumps it (``_colors.scss`` is where the grayscale property is set) due to it being loaded after our PL site-centric styles (where we are defining the true black). 

I didn't think adding an ``!important`` was crucial for this case, but could go either way since other warnings/messages on the PL might need to trump pattern styling in the doc site's case.

Your call, really.

clrux:
No worries. I prefer it without the flag too.

PR #43, opened 2015-04-10
Fixup - Adjusting example set to be reusable

clrux:
@talbs This syncs up the example wrappers for our color palettes.


talbs:
:+1: 

Mind changing your commit message to remove the "Fixup -"? We usually use that prefix or something like it (i.e. "fixup!") to signify commits should be squashed. Also, mind prefixing the commit with a note that this changes the PL site and not any actual patterns?

talbs:
Also, it looks like you didn't check in the main.css file that is compiled with your selector change. I'm getting a local diff on that file and am not able to change branches locally.

talbs:
:+1:

PR #48, opened 2015-04-15
Changing links from # to edx.org

clrux:
@talbs This really quick PR changes the link/button `href` from `#` to `edx.org`.


talbs:
:+1:

PR #49, opened 2015-04-15
Changing links from # to edx.org

clrux:
@talbs This gets the link fixup to the site.

### DIFF
--- a/_posts/elements/2015-04-09-buttons.md
+++ b/_posts/elements/2015-04-09-buttons.md
@@ -98,8 +98,8 @@ info: Buttons should be used for performing actions within the edX environment. 
 
 <h3 class="hd-6 example-set-hd">Links</h3>
 <div class="example-set">
-    <a href="#">A normal link</a>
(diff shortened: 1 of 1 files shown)


talbs:
Does this have to go out now (vs being a part of an upcoming push)? If so, can you do the following?

* make sure its merged to master already
* rename this branch to start with ``clrux/publish-*`` (its easier to know what branches are for publishing this way)

talbs:
If it can wait, I'd really recommend that as our previous publish branches/PRs have just taken the most recent set of commits that are the diff between ``master`` and ``gh-pages``. This would break that as there are other commits before this that haven't been merged to ``gh-pages`` yet.

clrux:
It can wait. Maybe we can roll it in with the icons?

talbs:
Yeah, that's fine. If you want to push before then (to get it out), you can make a PR of against ``gh-pages`` from the tip of ``master``.

Mind closing this and either doing that or waiting till we have icons?

clrux:
Not necessary now. Will roll this update into the next publish branch.

PR #54, opened 2015-04-27
Colors: Simplification

talbs:
This work simplifies a lot of our color settings based on design and FED/Library management feedback. This includes the following efforts:

* removed mid-* values
* added in utility values
* simplified grayscale sets
* updated examples
* removed brand-accent setting/scale
* updated heading/typography color settings


talbs:
@clrux, mind reviewing?

clrux:
I was poking around being nosey and saw this. Loved it! :+1:

PR #57, opened 2015-04-28
Changing compiled CSS output to compressed

talbs:


_src/pl/static/sass/_components.scss
```diff
@@ -15,7 +15,7 @@
 .example-set {
     @extend %clear-last-child;
     padding-bottom: spacing-vertical(mid-large);
-    border-bottom: spacing-vertical(xx-small) solid palette(grayscale, xx-light);
+    border-bottom: spacing-vertical(x-small) solid palette(grayscale, xx-light);
```
talbs: NOTE: fixing Sass compilation error (``xx-small`` was not defined).

talbs:
@clrux, here's another quick PR to make sure our output CSS is minified (I had previously kept line breaks in settings).

clrux:
:+1:

PR #58, opened 2015-04-28
Fixup! - Addressing lint issues and script arrangement

clrux:
@talbs This addresses the:
* location of our 'vendor' scripts
* creation of a new gulp file for scripts not specific to the 'pl'
* lint errors within the existing JS files
* new JS addition to assets


gulp/tasks/scripts.js
```diff
@@ -0,0 +1,19 @@
...
+var concat          = require('gulp-concat');
+var rename          = require('gulp-rename');
+var uglify          = require('gulp-uglify');
+
+gulp.task('scripts', ['pl_scripts-lint'], function() {
+    return gulp.src([
+            // setup script sequence
+            './_src/vendor/svg4everybody/svg4everybody.min.js' // polyfill for SVGs in IE9-11
+        ])
+        .pipe(concat('vendor.js'))
```
talbs: Hmmm, eventually this will mirror the ``pl_scripts.js`` task and we won't be just concatenating vendor files but vendor files and then any custom work/files to one file. WIth that in mind, I'd prefer that we stick with the convention of having the concatenated file be named ``main.js``. In this case this would eventually be minified and placed in ``/public/js/main.min.js``.

Can you sync up that name?

gulp/tasks/scripts.js
```diff
@@ -0,0 +1,19 @@
...
+        .pipe(uglify())
+        .pipe(rename({suffix: '.min'}))
+        .pipe(gulp.dest(config.pl_local)) // move just for browersync + local preview
```
talbs: These should not have the PL documentation site config settings. It should have the pattern configuration (https://github.com/edx/ux-pattern-library/blob/master/gulp%2Fconfig.js#L51) instead.

gulp/tasks/default.js
```diff
@@ -4,7 +4,7 @@ var runSequence     = require('run-sequence');
 gulp.task( 'default', function() {
     runSequence(
         ['styles', 'images'],
-        ['pl_styles', 'pl_scripts', 'pl_images'],
+        ['pl_styles', 'pl_scripts', 'scripts', 'pl_images'],
```
talbs: NIT: move the ``scripts`` task to be within the rest of the general pattern tasks, like so:

```js
...
runSequence(
    ['styles', 'scripts', 'images'],
    ['pl_styles', 'pl_scripts', 'pl_images'],
...
```

talbs:
@clrux, left you a few notes. Let me know when you've addressed them so we can get this good fix in.

talbs:
:+1:

PR #59, opened 2015-04-30
PL: fix - correcting missing font-size Sass map value

talbs:


talbs:
@clrux, quick fix for a font-size Sass map value that I eliminated in my copy clean up. Mind reviewing?

clrux:
:+1:

PR #51, opened 2015-04-17
Elements: Icons (SVG)

clrux:
# Icons
This piece of work includes a new set of icons for use within the edX platform. Due to lack of SVG support in our previous library (Font Awesome) we decided to investigate other libraries that provided SVG filetypes.

The demo page has a range/slider control that lets users view the icons at various sizes.

**Requirements**
* find a font icon library that let us build a customized set,
* find a library that let us customize our set after we've built them,
* find an icon library that provided SVGs in addition to typeface/fonts and images, should they be needed

## Icomoon
I decided to go with [Icomoon](https://icomoon.io/app/#/select) because it meets all of the above requirements. It offers an intuitive interface that lets us pick the icons we want, customize them, rename them, assign CSS prefixes (if needed) and suffixes (if needed). It provides a convenient .json file which we can upload later to further customize our set.

**Included**
* SVG files
* PNG files (as backups should we ever need them)
* JSON file (configuration and customization info for later revisits)

---

@talbs I believe this is ready for your review!
@cptvitamin Would you mind reviewing too?

PR #64, opened 2015-05-06
Adding Font Awesome set and examples

clrux:
@talbs This PR removes the Material Design icons and replaces them with our current Font Awesome set from the platform. I've tried to use the matching icon in Material, with a few variances.

__Question:__

I originally placed the `font-awesome.css` file(s) in `vendor`, but realized our `gulp` script assumes JavaScript files here, so the icons weren't working. Nor were they being ported over correctly. I ended up just placing the files manually and calling it in the header manually.

How do I do this so that it's in the `gulp` build flow?


talbs:
@clrux, I think you misunderstood what @frrrances and I were explaining this AM. In https://openedx.atlassian.net/browse/UX-2134, I tried to explain that:

1. we want to keep the good icomoon SVG method of using icon libraries
2. but we need to stick with Font Awesome and have our initial set represent what specific icons we're using from that library in our apps (edx-platform)

With that said, **the older way of using Font Awesome is not what we want**. Here's what the ticket says again:

- - -

#### GOAL: 
There's a problem with the current implementation of icons within the PL (http://ux.edx.org/#icons) - they are not the icons used in the platform currently and need to represent them and ONLY them for the time being.

#### OUTCOME: 
A merged set of work that: 

1. catalogs what Font Awesome icons are used within the edx-platform repo's UI, 
2. updates the current IcoMoon-managed collection with just those (removing the incorrect Google Material Design icons, per https://github.com/edx/ux-pattern-library/pull/51#issuecomment-97402534)
3. updates the PL doc site icons example with rendering of all of these icons.

- - -

**TLDR - Use the magical and a11y-friendly method you added previous, but do it with a set of only the Font Awesome icons we're really using now, please.**

Let me know if that's still unclear.

clrux:
I get it now, sorry. Will have something today!

clrux:
Also, it's kind of hilarious to me that I misunderstood. Making the set from icomoon was like, the easy solution, lol.

PR #65, opened 2015-05-07
Adding Font Awesome fonts

clrux:
@talbs This PR removes the Material Design icons and replaces them with our current Font Awesome set from the platform.

Note: there's some funkiness with the sizes with the FontAwesome set for whatever reason. When all svg's are set to the same height/width, there are a handful that appear larger. I've checked the styles and markup and I think it must be the "drawings" themselves. Any thoughts here?


talbs:
@clrux, this looks damn good. Thanks for the re-work.

One thing to address before merging - let's keep our icon prefix library agnostic, meaning we shouldn't use the ``fa-*`` prefix on these icons (since down the road our collection might include a mix from various libraries and we'll have less use for the Font Awesome cheatsheet of icons as our PL site will become the new cheatsheet). 

Instead, **can you add back the generic prefix of ``icon-*`` to the configuration and the demo here**?

- - -

Once that is done, :+1:

clrux:
@talbs Updated the prefix. Mind taking one more spin before we merge?

talbs:
:+1: , thanks for making this usable right now.

PR #70, opened 2015-05-15
Simplifying font-face calls

talbs:
This work simplifies the number of calls to our typefaces that we load through the ``font-face()`` mixin so that we're only using the ones needed by the modern browsers we support (and removing the files we don't need any more).


talbs:
@clrux, mind reviewing this proactive work?

clrux:
Love the formatting :+1:

PR #71, opened 2015-05-17
PL Doc Site: Formally Consume edX Pattern Library + RTL UI Support

talbs:
This work moves the documentation site (ux.edx.org) to consume the edX pattern library like other apps will (through importing edX PL Sass partials in its main.scss compile).

I've also used [bi-app](https://github.com/anasnakawa/bi-app-sass) for any relevant properties within  the documentation site and in patterns themselves. I've created logic for a main-rtl and main-ltr CSS file (in prep for any future RTL support)

A few notes: 

* ~~There is currently a weird issue with certain Sass function calls being rendered as text strings for CSS values. I've looked into [Sass Interpolation](http://webdesign.tutsplus.com/tutorials/all-you-ever-need-to-know-about-sass-interpolation--cms-21375) some, but that didn't seem to help. This was also run into with the certificates work that used the edX PL in the same way in https://github.com/edx/edx-platform/pull/7877~~. - **Fixed common sense found in https://github.com/edx/ux-pattern-library/commit/5a67ea400ee87182f74fba0006b3a9ebd8323469**
* The Gulp watch workflow now only kicks off for any Sass changes made to the ``/pl`` directory Sass. We may need to adjust this to watch for pattern changes too.

- - -

**Reviewers**
* @clrux


talbs:
@clrux, this is pro-active and extra-curricular, so please review when you can, but don't let it get in the way of sprint commitments.

clrux:
@talbs Few questions, but otherwise :+1:. Wondering if the quotes in the variable/map calls is contributing to the string output?

PR #74, opened 2015-05-19
Publish: Fonts/Icons/UI fixes and Prep for Theming/Packaging

talbs:
This work publishes the following:

* fixes icons to use Font Awesome Library
* fixes overflow problem with documentation site sidebar
* preps pattern source code for packaging + converts documentation site to use packaging structure
* prep for theming overrides
* simplifies font-face calls to modern browser support

PR #76, opened 2015-06-02
PL: revising supplemental nav + element :target styling

talbs:
This work adjusts a few things about the PL doc site sidebar:

* remove the :target styling for any element that is the current target (since the susy grid moves element content away from being flush against the sidebar in large viewports)
* removes an extra stroke/border on the sidebar that is not needed and causes issues when the sidebar's overflow-y styling causes a scrollbar.


talbs:
@clrux, here's a fix for your good feedback on https://github.com/edx/ux-pattern-library/pull/74#issuecomment-103476599. Mind reviewing?

clrux:
:+1: Thanks!

PR #81, opened 2015-06-03
Adding HEX values to color swatches

clrux:
@talbs This piece of work adds HEX values to the swatches. Because the RGBa lines are sometimes long, I've made the meta boxes stack (`display: block`). This adds vertical space, but lets them read clearly. I'm not 100% happy with it (I'd still like to go to something similar to what [Google does](https://www.google.com/design/spec/style/color.html#color-color-palette)) and I'm open to suggestions for display.

---

cc @andywaldrop


andywaldrop:
thanks!

_src/edx-pl/sass/components/_colors.scss
```diff
@@ -61,13 +61,19 @@
 .swatch-meta {
     padding: spacing-vertical(xx-small) spacing-horizontal(small);
     white-space: nowrap;
-    font-size: font-size(small);
+    font-size: font-size(x-small);
     @include text-align(center);
     color: palette(grayscale, dark);
 
     .color-rgb {
```
talbs: NIT: can you abstract this stuff out and make a new rule for:

```scss
.color -rgb, .color-hex {
    display: block;
}
```

talbs: Please add a margin-bottom using ``spacing-vertical(___)`` to this.

_posts/elements/2015-04-01-colors.md
```diff
@@ -23,6 +23,7 @@ info: These are the predefined colors used throughout our patterns and elements.
             <span class="color-class lod">x-dark</span>
         </div>
         <div class="swatch-meta">
+            <span class="color-hex"></span>
```
talbs: NIT: please change the order so that RGB is above hex values.

_src/edx-pl/sass/components/_colors.scss
talbs: While we're here, let's make it easier for folks to copy these (a la the git branches in the github UI - screengrab below), I'd love to see us be able to double/triple click to select each RGB or hex value for easy copying.

easy-copying-github

Mind trying to add this as well?

talbs:
@clrux, aside from the NITs and ask for easier copying, this looks good. Once you tackle those I'm cool.

- - -

:+1:

_src/edx-pl/sass/components/_colors.scss
clrux: This is just normal behavior. The reason it works so nicely on Github is because there are no spaces. I'm going to try and remove the spaces between commas in our values. If I can get that working, it should do the trick.

PR #82, opened 2015-06-04
Adding RequireJS and modifying files to load thru it

clrux:

---

## Using RequireJS

### Immediate load
```
<script data-main="{{ site.baseurl }}public/pl/js/required.js" src="{{ site.baseurl }}public/js/require.js"></script>
```

### Dependencies
```
```

---

__Reviewers:__
Talbs for overall overallness
Alasdair for RequireJS work


clrux:
Closed and replaced by https://github.com/edx/ux-pattern-library/pull/85.

PR #83, opened 2015-06-04
Sync Up - Certs + Pattern Library

talbs:
This work syncs up any static changes made when using the Pattern Library in the recent Web Certificates UI (https://github.com/edx/edx-platform/pull/7882). This includes:

* revising ``edx-pl`` source directory to be named ``pattern-library``
* removing unnecessary Open Sans typeface file formats + weights
* updating README
* making the ``pl`` namespace/prefix a bit more clear by changing it to ``pldoc`` in HTML, styling, and workflow tasks

- - -

**Reviewers**

* [ ] FED - @frrrances 
* [x] FED 2 - @clrux

PR #84, opened 2015-06-05
Fix: Loose Ends from Dir/Naming Convention Changes

talbs:
This work cleans up after revising dir/naming conventions added in https://github.com/edx/ux-pattern-library/pull/83.


talbs:
@clrux, found a few things that I forgot in #83. Mind taking a quick look?

clrux:
Much of this is goin' the way o' the dodo after my work, but this is good 'till then. :+1:

talbs:
Thought your work was JS-specific, but I'll take your word for it.

PR #85, opened 2015-06-05
Adding RequireJS

clrux:
__Replaces https://github.com/edx/ux-pattern-library/pull/82__

---

---

## Using RequireJS

### Immediate load
```
<script data-main="{{ site.baseurl }}public/pl/js/required.js" src="{{ site.baseurl }}public/js/require.js"></script>
```

### Dependencies
```
```

---

__Reviewers:__
@talbs for overall overallness
@AlasdairSwan for RequireJS work and assistance


clrux:
@AlasdairSwan If you have a little time on Monday, would you mind helping me out a little with what's left? I'm stuck.

talbs:
@clrux, thanks for the work in getting our JS synced up with RequireJS and how things are done in the platform. I'll take a more detailed look through, but here are a few items for now:

* **Extra Bold Font Face Files** - It looks like you've added back in the ``_src/pattern-library/fonts/OpenSans/OpenSans-ExtraBold-webfont.*`` files with this work. These shouldn't be in the repo (as we are removing them from the platform and have already removed them from the PL repo + do not use them in our weight definitions). Please get rid of those.
* **Public pldoc Font Face Files** - Either manually or through Gulp tasks, you've added back in files within``public/pldoc/fonts/OpenSans/``. These fonts should be in the ``pattern-library/fonts`` directory both in the source and then moved to public (for Jekyll to build the ``_site`` directory with). Can you make sure any workflow/files you change address this?
* **Update README and Doc** - This is good work and the new way of using/including JS should be documented (and also point to any larger RequireJS documentation the ENG team and edX has). Can you update our README with this new software requirement and information on how/where/when its used alongside the larger doc reference/links?

I'll swing through in detail shortly. Thanks again!

clrux:
Replaced by https://github.com/edx/ux-pattern-library/pull/97

PR #86, opened 2015-06-11
Color swatch layout rework

clrux:
This work updates the layout of the color swatches to be more slim and streamlined, while maintaining at-a-glance use and values.

Some thoughts:
* removing the space and padding between swatches allows us to see how colors may look next to one another, which could be useful for UI design
* the wider swatches allows us to bring all of the relevant usage info on one line, making it easier to read and follow
* accessibility color highlights are now done as a left-border to the swatch itself for vertical scanning
* we should probably note somewhere __what__ the accessibility highlight means

---

@talbs Care to review this extracurricular work?


talbs:
@clrux, I've added a commit that revises some layout/visual bits and also simplifies the markup and styling used here. I've tried to sync up your JS logic where appropriate but it could use a double-check.

FYI, I've added in a nice touch that @frrrances suggested - giving folks the actual color code to use in Sass. I've left the instructional copy in the DOM but ``sr-only``'ed it since it may not be needed.

**Can you 1) rebase against master, 2) adjust the description of the colors to note that hex, rgb, and actual Sass color syntax are available now per color, and 2) proof what I've done?**

_posts/elements/2015-04-01-colors.md
```diff
@@ -13,441 +13,693 @@ slug:         colors
 
 description:  Pattern Library color swatches
 
-info: These are the predefined colors used throughout our patterns and elements. Values here (noted in RGB format as well by the relative scale name) are used for backgrounds, text color, and decorative elements. Note - when using any values here, make sure that any text on a background passes WCAG AA guidelines for color contrast.
+info: These are the predefined colors used throughout our patterns and elements. Values here (noted in RGB, HEX, and the Sass variable) are used for backgrounds, text color, and decorative elements. Note - when using any values here, make sure that any text on a background passes WCAG AA guidelines for color contrast.
```
talbs: NIT: mind changing this order to match the order we present them in within the UI (Sass Var, RGB, Hex)?

_src/pldoc/static/js/ui.js
```diff
@@ -110,20 +110,33 @@ $(document).ready(function() {
...
+            selection.removeAllRanges();
+            selection.addRange(range);
+        }
+    }
 
-        $this.mouseup(function() {
-            $this.unbind('mouseup');
-            return false;
-        });
+    $('.color-meta div').click(function() {
```
talbs: Any chance we can use the ``is-copyable`` class with this? I've added that class to all of the meta areas as well as a new place (color-reference). It'd be great to use a more utility-based class.

talbs:
:+1:

PR #87, opened 2015-06-11
FIX - Local Dev Configuration + Missing Assets Loaded into Views

talbs:
This works tackles a few things:

* PL Doc Site: moving redundant font face files to static public dir for site rendering only (this needs to be synced up only for the doc site use. All other apps will use the fonts found within ``pattern-library/fonts/`` + our Django Pipeline static collection
* Gulp: config paths + removing not used pldoc_styles-move.js task
* removing public asset artifacts from older directory structure (old ``/pl/`` stuff)
* removing local testing 'console.log' from color element JS logic


talbs:
@clrux, mind reviewing this summer cleaning list?

clrux:
The gulp tasks will be modified heavily by the RequireJS work. I _almost_ recommend you just do this there to keep it clean (you have my permission to push to that branch). I'm hesitant to make yet another branch that does restructuring because I'm afraid we'll be spending several hours rebasing and fixing conflicts. The `console.log` bit can come out as I'm taking care of that right now.

What do you think?

talbs:
@clrux, we talked about this in person. The order of operations for all PRs is:

1. https://github.com/edx/ux-pattern-library/pull/75 - assuming you can address feedback reasonably quickly
2. https://github.com/edx/ux-pattern-library/pull/87 - this one!
3. https://github.com/edx/ux-pattern-library/pull/86
4. https://github.com/edx/ux-pattern-library/pull/85

Are we still on the same page with that order? 

I'd like to get these fixes in sooner rather than later (as they affect any local dev work and will block any publishes to ux.edx.org.)

- - -

FYI, @frrrances.

clrux:
@talbs Yep, this order still works for me. I've got most of the forms stuff done, but I'm having some issues with `@extends` not being included. I'm gettin' a little stressed with it today so I'm going to pick it up on Monday and switch gears. Goal is to get the forms stuff merged Monday or Tuesday, then continue on with the rest of this list/order.

clrux:
:+1: once I merge in my Form stuff.

clrux:
@talbs Rebase versus Master then....

:airplane: this :lollipop: :house:

PR #90, opened 2015-06-18
add favicon

frrrances:
so i can find it in my tabs.

@talbs here you go.


talbs:
> As a @frrrances, I would like to have a fancy favicon so that when I have a ton of NYT tabs open, I can still easily get back to Pattern Library business

 :stuck_out_tongue_closed_eyes: 

- - -

Looks good. Thanks so much for the surprise and quick contribution.

:boom: :+1: :boom:

frrrances:
That use case is entirely valid.

PR #92, opened 2015-06-19
PL: Add Independent Views

talbs:
This work adds in the ability to create independent static pages/views under ``/examples/[viewname].html``. A few more details on this:

* these views use the ``pldoc/css/main-*.css`` files, not a separate stylesheet (as all PL documentation site styling should be scoped by namespace)
* there is a ``blank.html`` file included to be used as a starter for any views
* there is a ``kitchen-sink.html`` file included to proof any raw markup styling and separation of documentation from source styling


talbs:
@clrux and @frrrances, mind giving this a quick look? 

This won't create any specific patterns or new things (aside from the kitchen-sink and blank examples). Its more for things coming shortly, like the grid, layouts, and global nav/header/footer.

clrux:
:+1: after rebase. Do you think this should wait until the other four PRs we've discussed merged, or because it's really unrelated, can go in anytime?

PR #93, opened 2015-06-19
Accessibility fixups

clrux:
This work addresses a handful of accessibility issues uncovered during an overall assessment of the Pattern Library in its current state.

We want to frequently perform accessibility audits on the Pattern Library every so often to ensure we are adhering to best practices, web standards, and meeting WCAG AA guidelines. This was the first of many (to come).

## Patterns reviewed
This audit reviewed all of the elements we've built thus far, including:
* colors
* buttons
* copy and headings
* forms
* icons

Additionally, it audited the pattern library demonstration site as a whole. In general, it looked at:
* how the patterns work in a real-world scenario, including link color and contrast
* tabs and focus management
* screen reader only text and ARIA descriptors
* overall page/site navigability and landmark/region identification

## What was adjusted
After the audit, we revisited a few things. Some values were adjusted to pass requirements and some markup was added to assist those with visual and cognitive impairments. Some notable adjustments were:
* contrast was increased in disabled buttons; a "not allowed" cursor was added to the disabled extend to assist visually those buttons that are disabled versus those that aren't
* default link color was darkened slightly to provide enough contrast on gray
* tabs in tabbed containers are now buttons instead of links; additionally after activating a tab, focus is sent to the respective container
* each form input field that has a specific data format requirement received helper text that provides an example as to that format; when the placeholder text (currently used to provide example formatting) clears on data entry, those users with cognitive impairments will still be able to see an example

---

__Reviewers__:
- [x] @talbs FED and Sass
- [ ] @cptvitamin Accessibility


talbs:
@clrux, pushed up a commit that adjust headings/copy colors some. Check it out and squash when you're ready.

talbs:
@clrux, I added another commit that syncs up headings configuration variable names with how we treat text/copy (for easier remembering and consistency). Feel free to squash my previous commit (with the fixup! prefix), but keep this second commit in your PR's history.

talbs:
@clrux, let me know when you have the following things sorted out from our in-person talk:

* [ ] tab focus + viewport issue
* [ ] disabled states of all buttons firmed up (plus disabled link removed)

talbs:
@clrux, I'm seeing some weird behavior when changing element tabs on a local copy of the doc site. Here's a quick video - http://cl.ly/image/2r0H3O3c3T1c. Can you remove the jumping that is happening intermittently?

clrux:
@talbs Mind checking the tab, scroll, and focus once more? I think I've fixed it.

talbs:
@clrux, scrolling/switching tabs now looks good. But I am now seeing a focus ring on whole elements now (see screngrab below). Is this deliberate and needed?

screen shot 2015-06-25 at 10 59 40 am

I'd prefer to not have this, but am happy to include some ``:focus`` state styling if we need it for a11y reasons.

talbs:
@clrux, once you address that ``:focus`` issue, rebase against master, and squash your commits, :+1: from my perspective.

clrux:
@talbs It looks like I may have made this branch from one other than master as there are a lot of off conflicts. I squashed this work into a single commit, but I'm going to make a new branch from master and cherrypick. I will close this PR, open another, and request one more thumb from you formally.

clrux:
@talbs Been wrestling with this PR rebase all day and I finally realized what it's not workin'. I made this after I merged in my earlier RequireJS work, so the codebase is from that. We'll need to wrap up that PR first, then we can merge this one.

talbs:
:+1: everything seems to be rendering fine. Thanks!

_posts/elements/2015-04-09-buttons.md
```diff
@@ -46,7 +46,6 @@ info: Buttons should be used for performing actions within the edX environment.
     <button type="button" class="btn btn-link btn-base">Link Base</button>
     <button type="button" class="btn btn-link btn-large">Link Large</button>
     <button type="button" class="btn btn-link btn-small">Link Small</button>
-    <button type="button" class="btn btn-link btn-base" disabled>Link Disabled</button>
```
clrux: @cptvitamin We removed the disabled link, because links are either enabled and in the DOM or not in the DOM. Would you agree with this?

cptvitamin:
@clrux a couple of technicalities that are having an a11y impact:

* id=select_01_suc is used on two form fields.  I think you want to call the second one select_01_disabled.
* #iconFontSliderValue has no ```<label>``` element associated with it.  I think the value is a measure of pixels.  It might be helpful to display visually.

Here are a couple of other opportunities to improve a11y:

* There are 5 implicit landmark elements without aria-labels: ```<header>```, 3 ```<nav>``` elements and the ```<main>``` element.
* The first ```<h1>``` is really long.  I suggest getting the subtitle in the span out of the ```<h1>``` 
* I suggest adding a role="banner" to the ```<header>``` element with an aria-labelledby that references the <h1> (you'll need to add an ID to that)
* the ```<nav>``` elements all seem to have an ```<h2>``` as their first child.  Add an ID to those and then add an aria-labelledby to the ```<nav>``` elements that reference those HEadings.
* add an aria-label="Main" to the ```<main>``` element.
* I want to point out that their is an icon the precedes each link in the Documentation and About this Project navs  that visually indicate that the links go to external resources.  There is nothing to convey that to a screen reader user.  You may want to add SR text after the link text that says (external resource) or something similar (low priority BTW)
* The contrast of the copyright is abysmal, as in it appears as though I am trying to read at the bottom of an abyss where there is no light.
* I would do a contrast check on the entire page.  I think the color scheme used for the syntax/HTML highlighting has contrast issues.  Try using HTML_CodeSniffer to identify contrast issues.

clrux:
Alright @cptvitamin to keep things moving, I'm going to take care of the first two here since they are errors. I'll get this guy merged in and then so @talbs can continue with his bit. Then I'll make another branch and continue work on the other "not showstopping" fixes.

PR #96, opened 2015-06-22
FIX: correcting @include palette() values for color listings

talbs:
A quick clean up that's needed from some loose ends on https://github.com/edx/ux-pattern-library/pull/86


talbs:
@clrux, mind making this fixes things?

clrux:
:+1: 

Random question... now that we've added essentially a third copyable value, do we want them all to be inline?

talbs:
> Random question... now that we've added essentially a third copyable value, do we want them all to be inline?

I thought the inline-block property was essential for easier copying, but maybe its your JS that is doing that. Feel free to remove that style rule if its not needed.

PR #97, opened 2015-06-22
Adding and configuring RequireJS in the Pattern Library

clrux:

---

## Using RequireJS

This allows us to specify on a per-file basis, the requirements and dependencies for given JS files. For example, in `assets-secondary.html` rather than including the pattern library JS and the documentation site JS, we simply point to RequireJS and give it a `data` file to load (the required files).

In `required.js`, we define those files that need to load up-front (required) and then any dependencies.

In subsequent files that are loaded, __those__ files in turn have their own sets of required files and dependencies. Files are loaded on an as-needed basis, rather than all at once, ultimately saving download overhead, speeding up the site in general, and giving us control over which files are used where.

### Concatenation and minification

Since we need to load each file individually, we can no longer concatenate all of the JS files into single, minified versions. We can still minify them individually though. Instead of bundling all of the files together and minifying them, we're only loading the JS files as they're needed, using their minified versions.

### Immediate load
```
<script data-main="{{ site.baseurl }}public/pldoc/js/required.js" src="{{ site.baseurl }}public/js/require.js"></script>
```

### Dependencies
```
```

---

__Reviewers:__
@talbs for overall cleanliness and functionality; make sure things work as expected
@AlasdairSwan for RequireJS work and assistance


talbs:
@clrux, proofing things visually I found the following: 

**Enhanced select fields aren't rendering properly**
screen shot 2015-06-22 at 3 56 48 pm


**Icon Demo BGs are missing**
screen shot 2015-06-22 at 3 57 47 pm

clrux:
Ahhh those rolled in after I started the RequireJS. Need to be added, good catch @talbs thanks. Standby...

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "edx-ux-pattern-library",
-  "version": "0.2.0",
+  "version": "0.1.46",
```
talbs: Jigga what? Are we going back in time in version numbers for our own UX PL? Also, we should figure out when to set these.

talbs: If this wasn't done on purpose, please revert this.

talbs: Double also, this file should be managed by your local node. Did you do a ``npm install --save-dev`` or a ``npm uninstall --save-dev``?

package.json
```diff
@@ -12,14 +12,12 @@
...
     "gulp-autoprefixer": "latest",
     "gulp-changed": "latest",
-    "gulp-concat": "latest",
     "gulp-filesize": "0.0.6",
     "gulp-filter": "latest",
     "gulp-imagemin": "latest",
     "gulp-jshint": "latest",
     "gulp-minify-css": "latest",
     "gulp-notify": "latest",
-    "gulp-rename": "latest",
```
talbs: Can you confirm that we don't reference this and that any other node packages don't?

public/pldoc/js/ui.js
```diff
@@ -0,0 +1,110 @@
+define([
```
talbs: What's the difference between this file and `` _src/pldoc/static/js/ui.js`` (https://github.com/edx/ux-pattern-library/pull/97/files#diff-1a34c101d67f2fe73266b3f7ecb8f768)?

talbs:
@clrux, left you a few comments and issues. Mind addressing them? Also, does any of the documentation in either our README file or our general styleguides need to be updated (to reference vendor/dependencies as well as gulp tasks)? If so, please update them and leave me a comment here with a link to any updates.

package.json
clrux: I'm trying to version this properly: `major.minor.fixup`. I guess it doesn't __really__ matter though. What do you think? If we're not going to use it, should we just remove it?

package.json
clrux: No errors are thrown when `gulp`ing so I don't think it's being used anywhere else.

public/pldoc/js/ui.js
clrux: Not sure. The one in `public` is made automatically, no?

clrux:
@talbs The `README` is good. There's nothing specific required that isn't in the pull down from Git. As for styleguides, we don't yet have one for JavaScript so there's nothing to update. However, since we included our first Pattern Library script (the select replacements), it might be wise to start one. What do you think? Is that too far in the rabbit hole at this point in time?

package.json
talbs: I think remove it for this PR. We should think about how and when to version, but let's not start here or have this slow this getting in. I'll ping ENG for any standards they have when versioning edX software. Thanks!

talbs:
> @talbs The README is good. There's nothing specific required that isn't in the pull down from Git. As for styleguides, we don't yet have one for JavaScript so there's nothing to update. However, since we included our first Pattern Library script (the select replacements), it might be wise to start one. What do you think? Is that too far in the rabbit hole at this point in time?

@clrux, The JS styleguide would be good to have, but let's not start it with this. I'm assuming that styleguide would borrow a lot from practices that @andy-armstrong, @dsjen, and @AlasdairSwan have put into place and documented elsewhere. 

**Would you mind making a UX PL story in our backlog that starts that styleguide (including gathering input/doc from the fellas mentioned above)?**

talbs:
@clrux, :+1: once you reset that versioning change.

talbs:
Woah woah, did @AlasdairSwan review and approve this?

AlasdairSwan:
Not yet @talbs due to Q4's rapidly approaching demise.

clrux:
Woops! Got a little too excited!

_src/pldoc/static/js/ui.js
```diff
@@ -1,142 +1,111 @@
...
+            var doc = document,
+                element = that[0];
+
+            if (doc.body.createTextRange) {
+                var range = doc.body.createTextRange();
+                range.moveToElementText(element);
+                range.select();
+            } else if (window.getSelection) {
+                var selection = window.getSelection();
+                var range = doc.createRange();
```
dsjen: FYI, range is declared twice essentially.  Once here and once on line 91.  JS is functionally scoped and the variables are hoisted to the top of the function.

PR #101, opened 2015-06-24
Using jQuery instead of scrolling plugin

clrux:
@talbs This very quick piece of work removes the `smoothScroll` plugin and relies on built-in jQuery `scrollTo` instead, reducing the number of vendor dependencies and page load overhead.


_src/pldoc/static/js/ui.js
```diff
@@ -12,14 +12,13 @@ $(document).ready(function() {
...
-            offset: -200,
-            easing: 'swing',
-            preventDefault: false,
-            speed: 1000,
-            scrollElement: null,
-            scrollTarget: $(this).attr('href')
+        $('html, body').stop().animate({
+            'scrollTop': $(target).offset().top
+        }, 1000, 'swing', function() {
+            $(target).find('.pldoc-heading').attr('tabindex', '-1').focus();
```
talbs: Should this be its own function outside of the ``smoothScrollLink()``  function you're borrowing here?

talbs:
@clrux, a few thoughts:

* I'm not sure the plain old anchor to an HTML id attr is working with this change. See this video of testing nav + the back to top link out - http://cl.ly/image/0F0e3Y2Z3u3h (I'll make a separate quick fix to adjust that back to top link's styling).
* I'd love a very very quick review of your JS from @dsjen or @AlasdairSwan to make sure this is how we want to roll this simple and homegrown function.

_src/pldoc/static/js/ui.js
```diff
@@ -12,14 +12,13 @@ $(document).ready(function() {
 
     // smoothscroll to target links
     function smoothScrollLink(e) {
+        var this = e.target,
```
dsjen: Renaming `this` makes me nervous.  It seems like you could get away with `$(e.target).attr('href');` rather.

_src/pldoc/static/js/ui.js
```diff
@@ -12,14 +12,13 @@ $(document).ready(function() {
...
 
-        $.smoothScroll({
-            offset: -200,
-            easing: 'swing',
-            preventDefault: false,
-            speed: 1000,
-            scrollElement: null,
-            scrollTarget: $(this).attr('href')
+        $('html, body').stop().animate({
+            'scrollTop': $(target).offset().top
```
dsjen: Nit: It looks like the convention in this file is to not use quotes for field names.  Should this be `scrollTop: $(target).offset().top` instead?

_src/pldoc/static/js/ui.js
clrux: Couldn't hurt, I guess. It would allow it to be referenced elsewhere if need be. Thanks.

clrux: Good idea, and it'll make that functionality available elsewhere, if it's ever needed.

dsjen:
The solution is small, and while I haven't tested, looks like it would work.  The only idea that I would suggest is that if you'd like this to be a general solution across pages, then moving that piece of functionality into a new component and "requiring" it in ui.js would be more modular.

At the same time, if that's outside the scope of what you were thinking or wanting for this change, it may not be worth it.

talbs:
@clrux, now that all of the RequireJS and larger bits have landed, do you want to revisit this PR?

clrux:
Yesir, thanks.

clrux:
So much has changed since I started this PR. It's quick and easy enough that I'm just going to it fresh rather than sift through commits fixing rebase conflicts. Closing this PR in lieu of another.

PR #102, opened 2015-06-25
Removing title from SVGs, adding sr-only

clrux:
This isolated work removes the `title` from the inline SVG as well as the SVG definitions file.

During accessibility testing, the `title` attributes were causing the SVG icons to be read more than once or not at all. We decided to use robust, accompanying `sr-only` text inline with icon fonts rather than using the SVG `title`.

One advantage to using inline `sr-only` text is that we can customize what each icon means individually, rather than having one meaning for that icon across the board.

__Caveats__

Unfortunately, whenever we update our set of icons, the `title`'s will re-appear since they come with the generated set of icon fonts.

---

@talbs No rush on this, but maybe before we make the next public/site update. Would be interested in hearing your thoughts on how to deal with the re-appearance of `title`'s whenever we re-generate the icon fonts.

@cptvitamin Question: would it be bad to have a `title` attribute in the SVG definitions file _and_ inline `sr-only` text?


talbs:
@clrux, this makes sense. :+1: 

* I think it will help maintenance of references to icons (its one less manual element to customize per reference), which will help the larger concern @andy-armstrong had.
* It is an extra step to strip out the title values from the data we get from our icon mgmt app, but I'm okay with that assuming that 1) it will be our team updating and curating that icon library and 2) we won't be doing it too often at all.

- - -

**Can you be sure to include this nuance to managing our set in your styleguide for icons?** It'd be good to have instructions for anyone on our team needing to update/add to the library.

cptvitamin:
@clrux no, that would be fine.  Screen readers do not read title attributes (unless the user goes *way* out of their way to configure their SR to do so, at the expense of *not* reading alts. The only use case for this is when a SR user has a job that uses software everyday where the devs added title attributes everywhere and no alts.)  At least one of our automated checkers actually flags title attributes as a potential error.  The only time I ever recommend using title attributes is on iframes, where they are critical and screen readers do read them (because there is no alt on iframes)

clrux:
@talbs Based on @cptvitamin's feedback, it's okay to keep the `title`'s in the generated SVG definitions file. So we only need to worry about the `title` attribute when actually using the icons.

I've included markup in the newly created (and yet to be curated) [Icon and icon font styleguide](https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Icons-and-icon-fonts). Whenever you get to reviewing that, let me know if it makes sense and anything we should add.

talbs:
@clrux, thanks for this extra work. The quick icons element styleguide looks great and has just the right amount of information. Its also good to note that we can keep the title metadata in the SVG files output by icomoon (keeping our maintenance of that speedy).

One last ask, can you add a note in the styleguide about how our icon fonts are managed (what service and by whom - both for edX and our recommendations for Open edX instance owners)?

:+1:

clrux:
@talbs I added a brief section to our styleguide about Icomoon and our custom set. Would you mind reviewing real quick? I'll merge this in, in the meantime, but let me know if there are any edits to be made.

PR #106, opened 2015-06-29
Set Up: Use Bower for Vendor/External File Maintenance

talbs:
This work moves all vendor/external files to be maintained through Bower (http://bower.io), a dedicated package manager on top of NodeJS. 

##### A Few Notes

* edX ENG already uses Bower in Insights (cc @dsjen) and other products
* Bower configuration has been created with both https://github.com/edx/ux-pattern-library/blob/talbs/use-bower/.bowerrc (setting where components are installed) and https://github.com/edx/ux-pattern-library/blob/talbs/use-bower/bower.json (similar to our ``package.json``)
* All existing static vendor file versions have removed while installing their replacements through Bower
* There is no change to how or where the vendor files are referenced
* All paths to vendor file referenced have been synced
* README and set up instructions have been updated

- - -

##### Reviwers

* [ ] General FED/references - @clrux 
* [x] Bower configuration and usage - @dsjen 

- - -

Remember to check out the new README file for instructions - https://github.com/edx/ux-pattern-library/blob/talbs/use-bower/README.md


talbs:
@clrux, heads up, this PR is extracurricular in nature, so it shouldn't get in your way of getting any sprint PRs merged. I'm most curious about any overlap this work has with your RequireJS work at https://github.com/edx/ux-pattern-library/pull/99. Will this be a large pain to integrate at around the same time?

clrux:
@talbs We have several open PRs at the moment, all of which futz with the structure to some degree. I recommend holding off on this one until those are wrapped up. Rebasing keeps getting harder and harder.

dsjen:
Nice!  Bower usage looks right to me. :+1:

clrux:
@talbs So the order of PRs that need to be merged are:
* RequireJS (waiting on review from @AlasdairSwan)
* a11y fixups
* this one

The second and third items here may be interchangeable. But I want to get RequireJS in first. Cool?

talbs:
@dsjen @clrux, this PR has been rebased to pick up the RequireJS (#108) work. I've also taken those new vendor files and moved them into using the ``requirejs-plugins`` Bower package. All of our vendor files continue to be handled through Bower with this update.

Mind taking another look through and making sure this is good to merge?

dsjen:
@talbs does requirejs need to be removed from the source?  It seems to still be there, https://github.com/edx/ux-pattern-library/tree/talbs/use-bower/_src/vendor/requirejs.  I would have expected that `bower.json` would be added to `devDependencies` too.  I think I'm missing something here! :bulb:

talbs:
> @talbs does requirejs need to be removed from the source? It seems to still be there, https://github.com/edx/ux-pattern-library/tree/talbs/use-bower/_src/vendor/requirejs. 

@dsjen, sorry! I rebased interactively and forgot to force push. That dir/file should be zapped now.

talbs:
> I would have expected that bower.json would be added to devDependencies too. I think I'm missing something here! 

@dsjen, apologies, by ``devDependencies``, do you mean this file - https://github.com/edx/ux-pattern-library/blob/talbs/use-bower/package.json? That's the NodeJS list of packages to install.  Right now I have instructions on how to install bower globally through npm (see the README updates), should I remove those and just add this to the project's dependencies so it gets caught and installed with a ``npm install``? **Just typing that makes me think, "yes, @talbs, you goof - of course we should have this be a project thing"** Sound good?

dsjen:
Oops, I mistyped in fact.  I was meaning that you'd add requirejs to `bower.json`, which I see is there now.  LGTM! :+1:

PR #108, opened 2015-06-30
Adding RequireJS

clrux:

---

## Using RequireJS

This allows us to specify on a per-file basis, the requirements and dependencies for given JS files. For example, in `assets-secondary.html` rather than including the pattern library JS and the documentation site JS, we simply point to RequireJS and give it a `data` file to load (the required files).

In `required.js`, we define those files that need to load up-front (required) and then any dependencies.

In subsequent files that are loaded, __those__ files in turn have their own sets of required files and dependencies. Files are loaded on an as-needed basis, rather than all at once, ultimately saving download overhead, speeding up the site in general, and giving us control over which files are used where.

### Concatenation and minification

Since we need to load each file individually, we can no longer concatenate all of the JS files into single, minified versions. We can still minify them individually though. Instead of bundling all of the files together and minifying them, we're only loading the JS files as they're needed, using their minified versions.

### Immediate load
```
<script data-main="{{ site.baseurl }}public/pldoc/js/required.js" src="{{ site.baseurl }}public/js/require.js"></script>
```

### Dependencies
```
```

---

__Reviewers:__
- [x] @talbs for overall cleanliness and functionality; make sure things work as expected
- [x] @AlasdairSwan for RequireJS work and assistance


talbs:
@clrux, this looks fine from both general PL code structure and local rendering perspectives. Looking forward to @AlasdairSwan (or someone else familiar with RequireJS if he's still swamped on Q4/Q1 stuff).

One last ask, can you update the README and any general general library mentions within our styleguide wiki to include the fact that we're using RequireJS.

:+1: once you add any needed documentation as well as the ENG review is complete.

_src/pldoc/static/js/required.js
```diff
@@ -0,0 +1,20 @@
...
+    }
+});
+
+require([
+    'jquery',
+    '/public/pldoc/js/ui.js',
+    '../../js/svg4everybody.min'
+    ],
+    function($, Ui) {}
+);
```
AlasdairSwan: Can we rename this file something more descriptive than "required"?

_src/pldoc/static/js/color-contrast.js
```diff
@@ -130,5 +132,4 @@ $(function() {
     };
 
     AccessibilityColorContrast.init();
-
 });
```
AlasdairSwan: I don't remember reviewing this file. There are some examples of 
```
var that = this;
```
which should be cleaned up and some variables not getting declared at the tops of functions.

AlasdairSwan: L87 don't need '=== true'

_src/pldoc/static/js/size-slider.js
```diff
@@ -55,5 +57,4 @@ $(function() {
     };
 
     IconFontSliderControl.init();
-
 });
```
AlasdairSwan: I didn't review this one yet either.

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,49 @@
...
+            hiddenClass:    'is-hidden',
+        },
+
+        init: function() {
+            Tabs.handleTabClick();
+        },
+
+        resetInterface: function(wrapper) {
+            wrapper.find('.pldoc-link').each(function() {
+                $(this).removeClass(Tabs.vars.activeClass);
```
AlasdairSwan: please replace $(this) with passed in value from function

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,49 @@
...
+            });
+
+            wrapper.find(Tabs.vars.panel).each(function() {
+                $(this).removeClass(Tabs.vars.activeClass).addClass(Tabs.vars.hiddenClass);
```
AlasdairSwan: please replace $(this) with passed in value from function
can you also please take a new line for .addClass

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,49 @@
...
+            });
+        },
+
+        handleTabClick: function() {
+            Tabs.vars.tab.find('.pldoc-link').on('click', function(e) {
+                e.preventDefault();
+
+                var content = $(this).attr('href');
+
```
AlasdairSwan: Please replace $(this) with $(e.currentTarget)

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,49 @@
...
+            wrapper.find(Tabs.vars.panel).each(function() {
+                $(this).removeClass(Tabs.vars.activeClass).addClass(Tabs.vars.hiddenClass);
+            });
+        },
+
+        handleTabClick: function() {
+            Tabs.vars.tab.find('.pldoc-link').on('click', function(e) {
+                e.preventDefault();
+
+                var content = $(this).attr('href');
```
AlasdairSwan: All variables need to be declared at the top of the function (because hoisting)

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,49 @@
...
+
+                Tabs.resetInterface($(this).closest(Tabs.vars.tabContainer));
+                Tabs.makeActive($(this), content);
+            });
+        },
+
+        makeActive: function(tab, content) {
+            tab.addClass(Tabs.vars.activeClass);
+            $(content).addClass(Tabs.vars.activeClass).removeClass(Tabs.vars.hiddenClass);
```
AlasdairSwan: new line for .removeClass please

_src/pldoc/static/js/ui.js
```diff
@@ -1,142 +1,111 @@
...
         }
-    }
+    };
 
-    $('.is-copyable').click(function() {
-        $(this).selectText();
-    });
-});
+    Ui.init();
+});
```
AlasdairSwan: Please replace all instances of this where possible

_src/pldoc/static/js/ui.js
```diff
@@ -1,142 +1,111 @@
...
-            var range = doc.body.createTextRange();
-            range.moveToElementText(element);
-            range.select();
-        } else if (window.getSelection) {
-            var selection = window.getSelection();
-            var range = doc.createRange();
-            range.selectNodeContents(element);
-            selection.removeAllRanges();
-            selection.addRange(range);
+        selectText: function(that) {
```
AlasdairSwan: Can you please rename the passed in variable to something more meaningful

_src/pldoc/static/js/ui.js
```diff
@@ -1,142 +1,111 @@
...
+            var doc = document,
+                element = that[0];
+
+            if (doc.body.createTextRange) {
+                var range = doc.body.createTextRange();
+                range.moveToElementText(element);
+                range.select();
+            } else if (window.getSelection) {
+                var selection = window.getSelection();
+                var range = doc.createRange();
```
AlasdairSwan: vars should be declared at the top of the function

_src/pldoc/static/js/ui.js
```diff
@@ -1,142 +1,111 @@
...
-            selection.removeAllRanges();
-            selection.addRange(range);
+        selectText: function(that) {
+            // this selects all the text in an element
+            // http://stackoverflow.com/questions/12243898/how-to-select-all-text-in-contenteditable-div
+            var doc = document,
+                element = that[0];
+
+            if (doc.body.createTextRange) {
+                var range = doc.body.createTextRange();
```
AlasdairSwan: vars should be declared at the top of the function

AlasdairSwan:
@clrux please clean up the gulp files so at the top there is only one var declaration

_src/pldoc/static/js/color-contrast.js
clrux: True, because if it's set, it is true. Thanks.

_src/pldoc/static/js/size-slider.js
clrux: I actually do need the object brackets. It fails to run without them.

_src/pldoc/static/js/tabs.js
clrux: `$(this)` is the jQuery object in each iteration of the loop. The `event` being passed is the array key (i.e., 0, 1, 2, 3, etc.). What did you have in mind here?

_src/pldoc/static/js/ui.js
clrux: Replace with something? Or remove?

clrux:
@AlasdairSwan I've gone through and addressed most of your feedback. I did have a few questions though, if you wouldn't mind taking another spin?

_src/pldoc/static/js/ui.js
AlasdairSwan: Use alternatives to
```
var that = this;
```

_src/pldoc/static/js/size-slider.js
AlasdairSwan: weird, that's fine 'tho.

_src/pldoc/static/js/ui.js
clrux: Oh, I thought you were talking about the specific line and was confused. Okay, on it.

AlasdairSwan: Sorry, I see the confusion. Apologies, I will use more words next time :)

_src/pldoc/static/js/color-contrast.js
```diff
@@ -36,20 +37,19 @@ $(function() {
...
+            var bg, fg;
 
             $('.example-color').each(function() {
 
-                bg = that.getL( that.rgbaToHex( $(this).find('.swatch-color').css('backgroundColor') ) );
-                fg = that.getL( that.rgbaToHex( $(this).find('.color-class').css('color') ) );
+                bg = AccessibilityColorContrast.getL( AccessibilityColorContrast.rgbaToHex( $(this).css('backgroundColor') ) );
+                fg = AccessibilityColorContrast.getL( AccessibilityColorContrast.rgbaToHex( $(this).find('.color-class').css('color') ) );
 
                 var ratio = (Math.max(bg, fg) + 0.05) / (Math.min(bg, fg) + 0.05),
```
AlasdairSwan: All variables should be declared at the top of the function

_src/pldoc/static/js/size-slider.js
```diff
@@ -22,38 +24,30 @@ $(function() {
...
+            IconFontSliderControl.vars.sliderControl.on('input change', function(event) {
                 // oninput - chrome, firefox, safari
                 // onchange - ie9-11
-                that.updateIconSize($(this).val());
-                that.updateAria($(this).val());
-                that.updateInputText($(this).val());
+                // console.log(event.target.value);
+                IconFontSliderControl.updateIconSize(event.target.value);
+                IconFontSliderControl.updateAria(event.target.value);
+                IconFontSliderControl.updateInputText(event.target.value);
```
AlasdairSwan: You use event.target.value 3 times so it would make sense to set it as a var at the top of the function and reference that instead.

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,54 @@
...
+            hiddenClass:    'is-hidden',
+        },
+
+        init: function() {
+            Tabs.handleTabClick();
+        },
+
+        resetInterface: function(wrapper) {
+            wrapper.find('.pldoc-link').each(function(event) {
+                $(this).removeClass(Tabs.vars.activeClass);
```
AlasdairSwan: replace this with event.currentTarget

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,54 @@
...
+            });
+
+            wrapper.find(Tabs.vars.panel).each(function() {
+                $(this).removeClass(Tabs.vars.activeClass)
```
AlasdairSwan: replace this with value passed into each function

_src/pldoc/static/js/tabs.js
```diff
@@ -0,0 +1,54 @@
...
+                       .addClass(Tabs.vars.hiddenClass);
+            });
+        },
+
+        handleTabClick: function() {
+            var el, content;
```
AlasdairSwan: these vars are only used within the .on function so their declaration should be moved there.

AlasdairSwan:
@clrux I reviewed and there are 3 files that need minor changes. Also, a note on consistency. For passed in events you sometimes use 'e' and sometimes 'event', for ease of reading could you please always use the full 'event' variables. When assigning jQuery object to variables it would be good if you could always prefix the var with $.

Once these changes are made we are good to go :+1:

clrux:
Thanks @AlasdairSwan. I've moved the variables, replaced `this` with `event.currentTarget` and went through and changed all `e` to `event` to be consistent.

clrux:
Merging with thumbs from both @AlasdairSwan and @talbs.

talbs:
FYI, @andy-armstrong

PR #109, opened 2015-06-30
FIX: Correct Path so Favicon Displays on All PL Doc Views

talbs:
This work sets favicon path to be site/root-relative so pages like http://ux.edx.org/elements/buttons/ can render it as well.


talbs:
@frrrances, here's a quick fix to a favicon issue, mind making sure it hits the mark?

frrrances:
first pr... fail! thanks for the fixup. :+1:

PR #111, opened 2015-07-02
FIX - PL: Addressing Local JS Lint Errors

talbs:
This work corrects the following output errors:

```
[17:00:00] Starting 'pldoc_scripts-lint'...
/Users/btalbot/edX/ux-pattern-library/_src/pldoc/static/js/color-contrast.js: line 48, col 37, Expected an assignment or function call and instead saw an expression.

1 error
/Users/btalbot/edX/ux-pattern-library/_src/pldoc/static/js/ui.js: line 82, col 20, 'hex' used out of scope.

1 error
[17:00:01] Finished 'pldoc_scripts-lint' after 141 ms
```


talbs:
@clrux, here's the quick fix for the local lint issues.

clrux:
Looks sound to me :+1:

PR #113, opened 2015-07-07
Set Up: Watch All Jekyll Files

talbs:
This work changes our local gulp workflow's watch task to watch more Jekyll-based file changes, including:

* HTML/DOM includes
* Layouts
* Example files (in ``/examples/*.html``)

This work should make any local template or copy changes much more efficient.


talbs:
@clrux, mind reviewing this quick fix to make us more productive when editing HTML/templates locally?

clrux:
:+1:

PR #115, opened 2015-07-08
Copy & Headings: changes to preview tab with labeling and example copy length

marcotuts:
**Description**
A quick change to add in supporting text labels for the copy and header section preview tabs to expose the underlying class names on this view (in support of the markup tab's role). Additionally, a shift to using one paragraph of copy to convey the styling as opposed to two. 

**Checklist**
- [x] review / check if I should be excluding .css files from this PR
- [x] check whether the file name should change to reflect "last updated" date. perhaps this is done automatically, I'm not sure. 
- [x] Q: would this be also helpful on the paragraph tags within the headings section? I realized it has a similar pattern / layout. 

**Screenshot with proposed changes**
Copy Section:
image

Headings Section:
image

**Reviewers**
@talbs


talbs:
@marcotuts, thanks for this! Here are some thoughts inline:

> review / check if I should be excluding .css files from this PR

Hmm, I don't think those should have changed (sometimes this happens either because of branch changes while gulp is still watching/compiling files or due to mysterious unknown cases). I'd remove them from the commit since you're just changing copy.

- - -

> check whether the file name should change to reflect "last updated" date. perhaps this is done automatically, I'm not sure.

I think the date is actually pulled from the post's metadata (or what Jekyll calls "Front Matter". This should be the "date: ______" bit at the top of the file. Updating this to reflect your good changes being the last update makes sense.

- - -

> Q: would this be also helpful on the paragraph tags within the headings section? I realized it has a similar pattern / layout.

Yes, this sounds good too. Also, Mind adding the same suffixes to the headings themselves in that post (e.g. (hd-1))?

- - -

**Also, don't forget to update the README to make yourself a contributor to the repo!**

- - -

:+1: once all of those small things are synced up. Thanks for the help!

_posts/elements/2015-03-25-headings.md
```diff
@@ -20,78 +20,78 @@ info: Headings are used to define important topics and section content into dige
...
-    <p class="copy copy-base">Lackadaisical fallibly next annoyingly shrugged and copiously inexhaustibly rode however peculiar more this spuriously flagrantly vigilantly crud more woolly effectively macaw among cardinal more alas darn near a conic primly more truly tarantula hello inventoried.</p>
+    <h4 class="hd-4 de-emphasized">The Fallen Legacy (de-emphasized h4)</h4>
+    <p class="copy copy-base">Lackadaisical fallibly next annoyingly shrugged and copiously inexhaustibly rode however peculiar more this spuriously flagrantly vigilantly crud more woolly effectively macaw among cardinal more alas darn near a conic primly more truly tarantula hello inventoried. (base copy)</p>
 </div>
 
 <div class="example-set">
-    <h5 class="hd-5">Diamond World</h5>
+    <h5 class="hd-5">Diamond World (h5)</h5>
     <p>Behind the hello far that that alas weasel vitally went one properly darn ladybug crud agonizing gosh so wicked goodness in until hedgehog some far close jeez some this hired dachshund bawdy frog outdid tedious after over but <a href="#">nonchalant ouch jeez thus</a> more some until capitally hen outside tarantula.</p>
```
marcotuts: Q: It looks like this example-set has no classes for its paragraphs. Is this intentional or should I add a specific class to the paragraphs here? (impacts entire example set)

_posts/elements/2015-03-25-headings.md
```diff
@@ -20,78 +20,78 @@ info: Headings are used to define important topics and section content into dige
...
 
 <div class="example-set">
-    <h6 class="hd-7">Slaves of Birch</h6>
-    <p class="copy copy-micro">Sourly browbeat yikes cantankerous jeepers far gosh crucially darn that so this the <a href="#">grumbled where jeez</a> and bid around the by warthog swung pending this stared hired and breezy stank bald a as some however dear since more more.</p>
+    <h6 class="hd-7">Slaves of Birch (h7)</h6>
+    <p class="copy copy-micro">Sourly browbeat yikes cantankerous jeepers far gosh crucially darn that so this the <a href="#">grumbled where jeez</a> and bid around the by warthog swung pending this stared hired and breezy stank bald a as some however dear since more more. (micro copy)</p>
 
-    <h5 class="hd-7 emphasized">The Heat's Game</h5>
-    <p class="copy copy-micro">Next stupid and on stuffily bald because domestically grew factious sewed outgrew coughed hello this overcame violent briefly much more improperly prior more experimental jeepers that egotistically licentiously unlike while darn regarding slowly mandrill until this dear on that mischievous before that insincere robin in nervelessly and.</p>
+    <h5 class="hd-7 emphasized">The Heat's Game (emphasized h7)</h5>
```
marcotuts: Q: Should this (and the next) heading be an h6 instead of and h5 or h4? I wasn't sure if the change in headings used in  the html was intended to show how you can apply the h7 styling to multiple levels of HTML elements (h4 --> h6) or if this was a typo.

marcotuts:
Thanks @talbs!  I've addressed the follow-ups but two questions came up while changing the headings file that I wanted to double check before squashing and merging. I've noted them in-context above.

_posts/elements/2015-03-25-headings.md
talbs: Hmm, headings were done before copy, so we didn't have those classes at the time. I think it makes sense to add copy-base to the larger ones and copy-meta to the smaller. Thanks!

talbs: Never mind saw you already added those in. Looks good!

_posts/elements/2015-03-25-headings.md
talbs: Yeah, we ran out of HTML level headings and it was a good opportunity to show that semantics (choosing the right heading for your DOM) can be separated from the presentation and styling. Unless it's confusing or baaaaad in your opinion, we can keep as is.

talbs:
@marcotuts, thanks for the thoughtful questions. I've added some thoughts to each. No action seems to be needed.

You still have my :+1: and things look good!

PR #116, opened 2015-07-09
Fixing disabled on $.clone

clrux:
@talbs This is a quick fix for the disabled states on replaced selects.


_src/pattern-library/js/select-replace.js
```diff
@@ -51,6 +51,7 @@ define([
                     }
 
                     if ($el.is(':disabled')) {
+                        console.log(replaced);
```
talbs: Mind removing your debug code here for production?

talbs:
@clrux, making your commit message a bit more descriptive (like what UI element and interaction does this fix)?

clrux:
@talbs Updated, sir!

talbs:
:+1:

PR #122, opened 2015-07-14
Removing scroll plugin in lieu of jQuery

clrux:
@talbs This fresh piece of work removes the scrolling plugin in lieu of using just plain-Jane jQuery.


talbs:
:+1:

PR #123, opened 2015-07-14
Added basic support for installation as a bower dependency

Nickersoft:
The library now allows for a customizable fonts directory, as well as changing all dev dependencies to dependencies. To use, install the git repo using `bower install`, then copy main_ltr.scss or main_rtl.scss to your project and adjust the dependency paths. Override `$ux-pattern-library-font-path` to use a custom font path.

@clintonb @talbs


clintonb:
FYI @clrux

clintonb:
:+1:

_src/pattern-library/sass/utilities/_variables.scss
```diff
@@ -141,55 +141,58 @@ $shadow: transparentize(palette(grayscale, black-t), .5);
 // ----------------------------
 // #TYPOGRAPHY
 // ----------------------------
+
+$ux-pattern-library-font-path: '../fonts' !default;
```
talbs: Would you mind changing this variable name to match some of the conventions used for the rest of our typography variables? Something like ``$font-path`` seems more appropriate.

talbs:
@Nickersoft, thanks for the pro-active contribution. Its much appreciated. I had one small NIT about the font path variable name you used, but other than that, this looks good.

:+1: once you address that.

_src/pattern-library/sass/utilities/_variables.scss
Nickersoft: @clintonb, thoughts?

clintonb: @Nickersoft you asked for my thoughts and almost immediately merged.

I suggested using a prefixed name here so as not to conflict with any other user variables, or the variables of other libraries. For example, FontAwesome uses the `fa-` prefix for most of its variables. Had Dave chosen to use `$font-path` instead of `$fa-font-path`, we would have a problem with integrating these two libraries. I *highly* recommend adding a prefix to avoid this problem down the road.

clintonb:
@Nickersoft bower.json needs to be updated with an actual version number.

Nickersoft:
@clintonb Sorry, I seemed to have overlooked that.. this is version 0.0.1 I take it?

clintonb:
Yes. Also, see my comment on the font-path variable.

Nickersoft:
@clintonb I see that now. I merged after receiving a thumbs up from both of you. In terms of recommended course of action, instead of opening another PR for reverting and then another to push the additional changes, should I just open another with these two changes included? I'll wait for Brian to address your comment before taking further action.

clintonb:
New PR with both changes. We'll move the conversation there.

-- 
Clinton Blackburn

_src/pattern-library/sass/utilities/_variables.scss
talbs: @clintonb I *disagree* with the prefix. The Pattern Library HTML/Sass code is not intended to be used with Bootstrap nor any other UI Frameworks that handle the same level of rendering as it does.

What you folks are doing with Oscar and Otto is not the intended use of the code in this repo and I don't think this is the time to inject exceptions or special cases based on that work. If you need to handle conflicts within your project, I'd recommend that you do that within your project's codebase and Sass compilation.

clintonb: I understand it is not the intended use case; however, until someone recommends a better implementation, it is the world we live in. Allowing the pattern library to be used free of potential conflicts eases the burden on our team as well as every other team that will need to consider a similar integration as we continue to break apart edx-platform. We are talking about a single variable, not an exception to the entire project. There is no slippery slope here.

clrux: @clintonb, this repo is meant to be clean and pure. We can't really allow exceptions like this, because that opens doors for more exceptions and then we're left with a heavily modified, no-longer-pure library. We've tried really hard to make this library extensible and easy to be themed and are available to answer any questions our documentation doesn't.

As @talbs said, this library isn't really to be merged with anything else, rather be a standalone library with which to build a platform or site. The work you guys are doing is a unique case, and we advise making any unique modifications to the project, rather than to this library.

Does this make sense?

PR #127, opened 2015-07-22
FIX: select-replace.js

talbs:
This work commits a compiled, but locally orphaned JS file.


talbs:
@clrux, here's a quick commit to get that JS file that's being compiled locally (with a diff) up to date. Seem okay?

clrux:
:+1:

PR #131, opened 2015-07-22
Elements: UI Depth

talbs:
This work, adds in a set of basic scale of visual depths to use through a page's layout, including:

* default page/view background
* -1 to -3 depth levels "below" the page/view background
* 1 to 3 depth levels "above" the page/view background
* Sass extends + utility classes for each level
* Sass extends + utility classes for UI-focused keywords (slat, card, dialogue, well, etc.)

- - -

##### Reviewers
* [ ] FED - @clrux  
* [ ] Visual Design - @andywaldrop


clrux:
@talbs It looks like UI Depth Level 3 might not be working properly. It looks like 0 (zero). I'm also not a fan of these being on a different shade of gray, as it's kind of hard to really tell what is the 'well' and what is just the example container.

talbs:
>  It looks like UI Depth Level 3 might not be working properly. It looks like 0 (zero).

^ @clrux, ahh, yeah, that would be the depth for a dialogue/modal element, but we haven't defined the style and behavior for that yet. I'm going to remove it until we need/do it. Cool?

talbs:
> I'm also not a fan of these being on a different shade of gray, as it's kind of hard to really tell what is the 'well' and what is just the example container

^ @clrux, yeah, I wanted to render these elements with depth-0 as a background (to show what they would really look like if used in the wild on a view/page). Unfortunately, our current doc site styling puts all element previews on a depth-2 level. This was my attempt at setting that depth-0 context within there. If its confusing, I can just remove that wrapper/background and render it on depth-2 until we revisit the design of the doc site. 

What's your call? Leave it in with that in mind or fall back to the depth-2 bg of the elements?

clrux:
@talbs Is there any way to have the collective view have them without the gray and then the dedicated view can have them as they might live in the wild? That make sense?

talbs:
@clrux, sure does! I'll make an example page with that context right quick! Thanks!

talbs:
@clrux, thanks for the good feedback. I've made a couple of ``fixup!`` commits to do the following:

* remove the depth-3/dialogue level of depth until we're ready to style/build modals
* created a new example page for depth (that shows the elements on a ``depth-0`` background - http://localhost:3000/examples/depth.html) and removed the ``depth-0`` visual context/background from the pattern listing at large (http://localhost:3000/elements/depth/)

Let me know what you think.

clrux:
@talbs Looks good, thanks. Only one more piece of feedback - is there any way to add a margin to the bottom of the example page?

After that, :+1:

talbs:
@clrux, margins added! Thanks much!

PR #133, opened 2015-07-23
FIX - PL: Adjusting Sass Compile Order/Organization

talbs:
This work does the following:

* adjusts compilation order for base partials to utilize components/elements.
* moves all simple "base" level elements within the ``pattern-library`` and ``pldoc`` source code to be contained under "utilities".


talbs:
@clrux, I found with the UI Depth work that I wanted to extend some of its styling into our base/reset styling, so this re-ordering of the Doc Site (and any other app's styling) compilation was needed. Also, found a redundant rule when I made this change.

Cool to change these bits?

clrux:
@talbs How does this affect other things? And do we still want to call that section "BASE" even though it's not really the base anymore, semantically.

talbs:
> How does this affect other things?

^ @clrux nothing else seems affected. All components/elements are scoped well with their extend/class-based selectors and base can now reference any of those.

> And do we still want to call that section "BASE" even though it's not really the base anymore, semantically?

^ Hmm, I think the purpose of that section is still **base styling for the app**, to me, we're just changing the order and loading elements/components. Elements/components don't inherit anything from the base styling, that section is merely for a quick reset/foundation for app. I can make that clearer in the doc. It may make sense to get more brains on this decision too and not rush it. @frrrances, thoughts?

clrux:
It might be as simple as calling it BASE APP then. BASE alone sounds foundational.

talbs:
True. That.

frrrances:
i like base app.

frrrances:
oh wait... this just changes the order. i think it's fine as-is with the new order.

talbs:
Yeah, true. Cool with just the re-order? If the confusion is that all of this compilation is app-centric, that can be addressed with better doc and overall comments/file naming separately.

frrrances:
:+1:

talbs:
@clrux, you cool with this line of thinking? In short, we'll be compiling each app's sass in this order:

* Vendor
* Utilities + Config (global settings - including RTL/LTR sensitive stuff)
* Components (Individual Patterns/elements)
* Base styling for the app
* Layouts
* Views
* Overrides

talbs:
Okay, after talking further with @clrux, we've decided to simplify even further. All "base" level elements (which really only contained more utilities like normalize, global/app style, and typography resets) have been moved under utilities.

@clrux and @frrrances, cool by you?

If so, I'll merge and update quickly any doc to reflect this simplification.

frrrances:
cool beans. :+1:

clrux:
:+1:

talbs:
Thx, gang. Doc has been adjusted - https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Sass-&-CSS#main-style-compile

PR #134, opened 2015-07-23
FIX: Remove Grid Debug Utility

talbs:
This work removes the debug grid extend/utility - it is not used and is causing compilation issues


talbs:
@clrux one more fallout from using an updated PL source code in the platform.

clrux:
:+1:

PR #137, opened 2015-07-28
PL Doc: removing '@include' syntax from color code copy

talbs:


talbs:
@clrux, mind reviewing this quick fix to correct the color syntax we include in those spiffy color range examples?

clrux:
:+1: 

Don't remember why we used `@include` before?

PR #138, opened 2015-07-28
revising jekyll paths to watch all files/dirs

talbs:


talbs:
@clrux, this fix makes sure the watchers will fire off tasks for all levels of Jekyll templates/includes (including folders within the main areas). Mind reviewing?

clrux:
:+1:

PR #140, opened 2015-07-29
Fixing scroll top issue

clrux:
@talbs This quick piece of work fixes the scroll top link and ensures focus is sent properly on nav or scroll-top click.


talbs:
:+1:

PR #145, opened 2015-08-12
FIX - Simplifying Sass baseline/spacing variables

talbs:
This work, based off of good feedback from @frrrances, simplifies how we use ``$baseline`` and the ``spacing-vertical/horizontal``settings in the PL. 

Essentially, ``$baseline-vertical`` and ``$baseline-horizontal`` are not needed and we can just scale vertical and horizontal spacing within the $spacing-`` sass maps.

- - -

##### Reviewers

* [x] FED - @frrrances


talbs:
@frrrances, mind reviewing this fix to your good catch?

marcotuts:
:+1:

frrrances:
:+1: Yay! Easy maths...

PR #148, opened 2015-08-28
FIX - PL: Correcting Icon Example Paths + Setting <base>

talbs:


talbs:
@clrux, this fixes your broken icon paths that are present currently on - http://ux.edx.org/elements/icons/. Mind, reviewing?

- - -

FYI, @frrrances

clrux:
I see what I did there.

:+1:

PR #154, opened 2015-08-31
Simplify edx-pattern-library Sass Compilation

talbs:
This work focuses on simplifying edx-pattern-library sass compilation syntax (similar to how its done in [Bootstrap Sass](https://github.com/twbs/bootstrap-sass#sass-1) to just one line when including it in an app's compilation. This involves:

* creating the ``_edx-pattern-library.scss`` partial to import all lib partials
* adding in per-app (pldoc) configuration overrides
* renaming/syncing comments and style headings

- - -

##### TODOs

* [x] Sync up Wiki and README Compilation code examples

- - -

##### Reviewers

* [x] FED + Sass architecture - @frrrances


talbs:
@frrrances, mind re-reviewing? I added some more simplification that let's anyone **call the edx-pattern-library in 1 Sass import**. Check out - https://github.com/edx/ux-pattern-library/commit/fc798d518e326c482746f69d8dde85880dc21c74.

In short, this:

* moves and third party library references that ``edx-pattern-library`` needs to the ``_edx-pattern-library.scss`` partial
* sets a ``layout-direction`` Sass var in the ``*-ltr/rtl.scss`` files
* that directional variable is then passed into the ``_edx-pattern-library.scss`` partial and any directionally-conditional third party libs (bi-app-sass) and variables (susy's grid-direction) are passed into the rest of the compile.

- - -

This means: 

* no more multiple paths to mess up when trying to add the edx-pattern-library to your app's Sass compile
* all libs are properly loaded with the edx-pattern-library automatically
* import paths are simplified once in the ``_edx-pattern-library.scss`` partial to just ``utilities/**/**``

- - -

I'm updating doc to accompany this now. FYI, @clrux and @marcotuts

_src/pldoc/static/sass/main-rtl.scss
```diff
@@ -6,17 +6,9 @@
 // NOTE: This is the left to right (RTL) configured style compile. It should mirror main-rtl w/ the exception of bi-app references.
```
frrrances: Note not needed anymore?

_src/pldoc/static/sass/main-ltr.scss
```diff
@@ -6,17 +6,9 @@
 // NOTE: This is the left to right (LTR) configured style compile. It should mirror main-rtl w/ the exception of bi-app references.
```
frrrances: Note not needed anymore?

_src/pldoc/static/sass/main-ltr.scss
```diff
@@ -6,17 +6,9 @@
...
 
 // ------------------------------
-// #VENDOR
+// #CONFIG - layout direction
```
frrrances: Are there other config things that might go in this section?

frrrances:
Totally sweet cleanup! Couple small comments, but otherwise :+1: :shipit:

_src/pldoc/static/sass/main-ltr.scss
talbs: Not in this section, but in the ``build.scss`` file, there is a place for the app's ``_config.scss`` which can add/replace variables, functions, mixins.

talbs:
Thanks, @frrrances. Removed the notes. Merging after I update the wiki docs.

PR #156, opened 2015-09-01
Add/Re-organize Path-Based Sass Variables

talbs:
This quick work adds in some quick configuration for static asset (fonts and images) path Sass variables in the edx-pattern-library.

- - -

##### Reviewers

* [x] FED/Abstraction - @clrux

### DIFF
--- a/_src/pattern-library/sass/utilities/_variables.scss
+++ b/_src/pattern-library/sass/utilities/_variables.scss
@@ -3,6 +3,7 @@
 
 // About: Sass partial for defining variables used throughout the applications.
 
+// #PATHS:          paths for particular assets
 // #UNITS:          Basic units of measurement
 // #GRID:           Grid and layout variables
 // #COLORS:         Base, palette and theme color definitions + application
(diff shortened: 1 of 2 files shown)


_src/pattern-library/sass/utilities/_variables.scss
```diff
@@ -14,6 +15,14 @@
 
 
 // ----------------------------
+// #PATHS
+// ----------------------------
+$static-path: '..' !default;
+$font-path: '#{$static-path}/fonts' !default;
+$image-path: '#{$static-path}/images' !default;
+
```
clrux: Are you making a separate variable for `$static-path` in case it changes? I feel like, at least like this, the "dot dot" as a variable is redundant and might be better included in the `$font-path` and `$image-path` variables.

talbs: Yeah, the thought might be that someone may keep their static assets in a different structure, so in that case, they could just change one variable rather than syncing two up. I can go either way on this. How strongly do you feel about simplifying it to 2 vars?

clrux: I'm 51/49 on changing it. I say that because, since this is our Pattern Library, I'd kind of like to keep it streamlined. If another person or organization needed to change the structure, I think we can safely assume they could easily create another variable as needed. Though, it is nice seeing the usage example on lines 21 and 22.

talbs: Good enough for me, I'll keep it simple and see how folks use it in their application of the UXPL. Thanks!

talbs:
@clrux, simplified the variables down. Mind reviewing what's there now?

clrux:
:+1:

PR #157, opened 2015-09-01
Add emphasized headings and copy extends

frrrances:
**Description**
This small PR adds extends to allow for emphasized copy and headings without needing to add classes to the DOM.

**Reviewers**
- [x] Code: @talbs

### DIFF
--- a/_src/pattern-library/sass/components/_copy.scss
+++ b/_src/pattern-library/sass/components/_copy.scss
@@ -9,17 +9,27 @@
 // ----------------------------
 // #GLOBAL
 // ----------------------------
+
+// broad stroke extend for emphasized/de-emphasized copy
+%copy-emphasized {
+    color: $text-emphasized-color;
+    font-weight: $text-emphasized-font-weight;
+}
+
+%copy-de-emphasized {
+    color: $text-de-emphasized-color;
+    font-weight: $text-de-emphasized-font-weight;
+}
+
 .copy {
     color: $text-base-color;
 
     &.emphasized {
-        color: $text-emphasized-color;
-        font-weight: $text-emphasized-font-weight;
+        @extend %copy-emphasized;
     }
 
     &.de-emphasized {
-        color: $text-de-emphasized-color;
-        font-weight: $text-de-emphasized-font-weight;
+        @extend %copy-de-emphasized;
     }
 }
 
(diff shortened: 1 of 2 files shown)


_src/pattern-library/sass/components/_copy.scss
```diff
@@ -108,4 +108,13 @@
     @extend %copy-micro;
 }
 
+// broad stroke extend for emphasized/de-emphasized copy
+%copy-emphasized {
```
talbs: Awesome addition! This will come in handy. Any chance I can ask for a further refactor that includes the following?

1. Define this near the top of the file (within the ``#GLOBAL``) section 
2. Use these extends to replace some of the [canned class rules here](https://github.com/edx/ux-pattern-library/blob/frances/emphasized-extends/_src/pattern-library/sass/components/_copy.scss#L15)

_src/pattern-library/sass/components/_copy.scss
```diff
@@ -108,4 +108,13 @@
...
+    color: $text-emphasized-color;
+    font-weight: $text-emphasized-font-weight;
+}
 
+%copy-de-emphasized {
```
talbs: Same as above, any chance I can ask for a further refactor that includes the following?

1. Define this near the top of the file (within the ``#GLOBAL``) section 
2. Use this extend to replace some of the [canned class rules here](https://github.com/edx/ux-pattern-library/blob/frances/emphasized-extends/_src/pattern-library/sass/components/_copy.scss#L20)

_src/pattern-library/sass/components/_headings.scss
```diff
@@ -138,3 +138,9 @@ $headings-count: 8;
         @extend %hd-#{$i};
     }
 }
+
+// broad stroke extend for emphasized heading
+%headings-emphasized {
```
talbs: Similar to the copy-based feedback, can you move this up to the ``#GLOBAL`` section of the headings partial and then extend it in place of the [canned rules here](https://github.com/edx/ux-pattern-library/blob/frances/emphasized-extends/_src/pattern-library/sass/components/_headings.scss#L28)?

_src/pattern-library/sass/components/_headings.scss
```diff
@@ -138,3 +138,9 @@ $headings-count: 8;
         @extend %hd-#{$i};
```
talbs: There's [a ``.de-emphasized`` class](https://github.com/edx/ux-pattern-library/blob/frances/emphasized-extends/_src/pattern-library/sass/components/_headings.scss#L33) that's applied at the global level in this file too. Should we make an extend for that (even though it just has a color-based rule associated with it).

talbs:
@frrrances, thanks for the good clean up and extra extends. I left a few small additional asks. Let me know if those make sense to add on.

Regardless, :+1:

frrrances:
@talbs I think I covered your feedback. Mind taking one more spin and letting me know if I got you loud and clear?

talbs:
:+1: 

That abstraction is da :bomb: !

PR #158, opened 2015-09-01
Adjust smallest breakpoint

frrrances:
**Description**
Adjust small breakpoint, and adjust margin on doc site for small screens

Before: 
screen shot 2015-09-01 at 4 28 53 pm

After:
screen shot 2015-09-01 at 4 28 56 pm

**Reviewers**
- [x] @talbs


talbs:
Thanks for the clean up, @frrrances. This looks good. Mind updating the version number here https://github.com/edx/ux-pattern-library/blob/frances/add-breakpoint/bower.json#L3 since this is a fix to the PL source code? 

I'm trying to follow the versioning convention (per ENG's guidelines) at http://semver.org/.

- - -

Once that's done as part of this commit, :+1:

PR #164, opened 2015-09-08
Adding styles for buttons with icons

clrux:
# WIP

These updates return to buttons and focus on icons within buttons.

## Solution

The current solution in progress instructs any SVG icons within a button to use the same dimensions as the button font size. It also has it use the same color as the font.

## Complications

Using SVGs within buttons ain't pretty, but it maintains accessibility and consistency with how we're using icons in the rest of the library. Additionally, we're specifying color/fill and size for icons in each button type.


talbs:
@clrux, should this be synced up with the work going on in https://github.com/edx/ux-pattern-library/pull/173 or just closed to be re-addressed once that's set?

clrux:
Let's close for now and re-open once the new icons are in place.

PR #165, opened 2015-09-09
PL Doc: Improve Local Dev Gulp Workflow

talbs:
This pro-active work cleans up a few things about the Pattern Library Doc Site's local gulp workflow, including:

* Sourcemaps for Sass/CSS (only in local development)
* Keeping local versions of compiled CSS to "expanded" view
* Adding in a production-ready task (with supporting cleaning and optimization tasks)
* Simplifying a lot of path references (in tasks + external asset calls) between Doc Site and Pattern Library source code
* Differentiating between Bower dev and general dependencies
* Adding JQuery as a Bower dependency
* Updating Node Packages to latest versions
* OCD: ordering gulp package vars in each task alphabetically
* Few bug fixes found along the way: Sass font variable path + missing semicolon in JS

- - -

#### Reviewers

* [x] @frrrances


talbs:
@frrrances, this is an extracurricular branch that got a bit carried away. In general this is all gulp-based tooling and re-organizing. In short, here are the highlights:

**Bower and Node Packages** - I updated our node packages to use the latest, split up dev vs. general Bower dependencies, and moved JQuery into our list of Bower reqs (vs. just having it statically here in the repo).

**pldoc_ vs. pattern-library pathing** - This was giving me headaches, so I: tried to change gulp tasks (and their names) to be simpler/more encompassing; set final output locations to be the PLDOC public dir; simplified any assets (syntax.css) I could to make it easier to get the two pipes processing similarly; and made some tasks more globally minded (script linting and image optimization).

**Sass/CSS Sourcemaps for local dev** - These seemed useful to have and some of the threads I was following with gulp package writers noted that the hang-ups to getting this implemented with the combination of Sass tools we use here (gulp-auto-prefixer, gulp-minify-css, gulp-sass, gulp-source-maps) were now gone. 

There was one hang-up, gulp-minify-css does not play well with the other three still in one task. So, I moved that into a separate task (styles-minify) that can be called by the ``production`` task (along with others) for when you're ready to commit/push the PL Doc Site's assets live. The silver lining is that we can now have expanded CSS output for easier debugging alongside the source maps locally and then have collapsed/minified output for production. The one change to workflow is that anyone who changes Sass/CSS should now run ``gulp production`` before making a commit :neutral_face:

^ That could be a pain, so if you'd prefer not having sourcemaps to running that command, I understand. I'm going to keep tinkering for a bit to see if I can get this worked out without that extra manual command.

**Bug Fixes** - I found a couple of bugs as I was working through this. They're in separate commits noted with "FIX:". I can move them out of this PR if that's helpful.

**OCD touches** - I re-organized some of the syntax of the gulp tasks to be similar across files (scripts, pldoc_scripts) and also sorted alphabetically gulp packages needed per task.

- - -

I'm looking to make sure that these changes make sense from a maintenance PoV and that they 1) do not fudge up any local workflows for you/us and 2) help with simplifying whatever we can.

talbs:
@frrrances, just an update on the issue in https://github.com/edx/ux-pattern-library/pull/165#issuecomment-141844018, I've got the styles task's packages (gulp-source-maps, gulp-autoprefixer, etc.) playing well with gulp-minify-css! So, no need to run any extra manual gulp commands before committing code, but I'm keeping the ``production`` task in there since its handy to prep all assets for launch. Cool?

talbs:
@frrrances, one last thing. When firing up this branch locally, make sure you update your local versions of the Node packages used in this project. a quick ``npm update`` should do the trick. If not, let me know.

Also, dev story created (https://openedx.atlassian.net/browse/UX-2908) for reviewing this work, punt it out of the sprint if you need to.

talbs:
Last update of the Sunday night, promise, @frrrances. 

I looked into why our Sass compilation/styles task is so slow to finish (for me its around ~24s on ``master`` now). After doing some digging and finding this good GH issue thread on Susy's repo (https://github.com/ericam/susy/issues/325), I wanted to see if that was the culprit for us. 

I added a "TEST" commit to this branch that removes susy from being called during the doc site's main-ltr/rtl.scss compilation. Here are the results when running the styles gulp task:

**This branch before TEST commit**
```
> gulp styles

Using gulpfile ~/edX/ux-pattern-library/gulpfile.js
Starting 'styles'...
Finished 'styles' after 31 s
```

**TEST commit**: 
```
> gulp styles

Using gulpfile ~/edX/ux-pattern-library/gulpfile.js
Starting 'styles'...
Finished 'styles' after 4.8 s
```

So it looks like Susy is what's causing the majority of the delay. :frowning:. I'll keep en eye on V3's release timing, which is supposed to include these optimizations. Hopefully we can pull it in fairly quickly.

- - -

(I'll remove the TEST commit before merging or when you want while reviewing)

talbs:
@frrrances, to help get this ready for review/merge, I removed the "TEST: does susy slow styles task down?" commit from this branch. You can still check it out and locally cherry-pick it if you want from https://github.com/edx/ux-pattern-library/commit/c8afaf4defe6ec38de27bc8b71e94708a34a422b.

Hope that helps!

frrrances:
Sounds like some good work! A couple of questions/comments from your comments above: 

- Does the JQuery req addition mean that JQuery will be installed with the PL or is it just in the dev dependencies? (Former might cause conflict with existing JQuery; Latter would be awesome!)
- Love the source maps and expanded css for development :)

talbs:
@frrrances,

> Does the JQuery req addition mean that JQuery will be installed with the PL or is it just in the dev dependencies? (Former might cause conflict with existing JQuery; Latter would be awesome!)

The ideal is to only JQuery here in the UXPL Doc Site, not when a person installs the UXPL bower package. With that said, I need to move that ``dependency`` into the ``devDependency`` list instead. Good catch! 

Making a fixup! commit now.

bower.json
```diff
@@ -40,12 +40,16 @@
     "package.json",
     "public"
   ],
+  "devDependencies": {
```
frrrances: So these additional dependencies that are only for dev?

talbs: Yeah, they should only show up when you run ``bower install`` from this repo. If you were in edX web certs or another place using this repo's bower package, they would not be installed alongside all of the items in the ``dependencies`` list. Nifty for local shiz.

gulp/config.js
```diff
@@ -17,70 +16,52 @@ module.exports                  = {
...
     styles:                     {
+
+        settings_development: {
+            outputStyle: 'expanded',
+            sourcemapsLocation: '.'
+        },
+
+        settings_production: {
+            outputStyle: 'compressed',
+            autoprefixer: { browsers: ['last 2 version'] }
```
frrrances: ooo. fancee.

bower.json
frrrances: :eyes: :gift:

pldoc/static/sass/_lib.scss
```diff
@@ -13,3 +13,5 @@
 } @else {
     @import '../../../bower_components/bi-app-sass/bi-app/bi-app-ltr';
 }
+
+@import 'syntax';
```
frrrances: What's this file do?

talbs: Ahh, its specific to the UXPL Doc Site needs (in this case, to show code). Its a CSS add-on for a Jekyll plugin called Pygments (see http://jekyllrb.com/docs/templates/#stylesheets-for-syntax-highlighting and the entire section above).

talbs: I was calling it in from a vendor file, but I've made some tweaks and I wanted to simplify the Sass/CSS folder structure, so importing it as a Sass partial made sense.

frrrances: Ah, v nice! Thanks for the explanation.

pattern-library/sass/utilities/_variables.scss
```diff
@@ -153,7 +153,7 @@ $shadow: transparentize(palette(grayscale, black-t), .5);
 // ----------------------------
 // typography: Open Sans
 
-$open-sans-font-path: '#{$font-path}/OpenSans/' !default;
+$open-sans-font-path: '#{$font-path}/OpenSans' !default;
```
frrrances: sweeeet.

frrrances:
Looks good @talbs ! Not an easy diff to follow but the changes look solid. I asked a few questions that it looks like you've already responded to so :+1: 

Would love to have a quick walkthrough of this setup and workflow for myself and any eng who might be interested...

talbs:
Thanks, @frrrances. I've squashed some fixup! commits and scheduled a walkthrough meeting with a larger gulp workflow gang.

PR #166, opened 2015-09-14
Fixing disabled dropdown behavior on replaced menus

clrux:
This work addresses an issue where any disabled JS replaced select menus would still function as if they weren't disabled. This was because the `$.clone()` function within jQuery doesn't capture the individual attributes so the cloned menu was enabled, despite it appearing disabled visually.

- [x] @talbs


pattern-library/sass/components/_forms.scss
```diff
@@ -292,6 +292,7 @@
 
     &:disabled,
     &.is-disabled {
+        cursor: not-allowed;
```
talbs: Y U NO use ``%state-disabled`` Sass extend here (https://github.com/edx/ux-pattern-library/blob/clrux/js-disabled-dropdown-fix/pattern-library%2Fsass%2Futilities%2F_helpers.scss#L13)? :P

talbs:
@clrux, I'm currently getting some errors on this branch locally when firing up the Doc Site.

* I'm seeing 2 JS errors in my console
* The enhanced select fields are not rendering properly (related to the first error I'm sure)

Here's a screengrab:
 
screen shot 2015-09-15 at 8 39 49 pm

Let me know when you have this resolved.

clrux:
@talbs Ah, I know why. This work requires the Gulp config fixes. I was baffled because I had this working. Should we merge that one first, then this one, or should I just combine them into a single PR since they are related?

talbs:
> @talbs Ah, I know why. This work requires the Gulp config fixes. I was baffled because I had this working. Should we merge that one first, then this one, or should I just combine them into a single PR since they are related?

Understood. Let's keep them separate and merge the other PR (https://github.com/edx/ux-pattern-library/pull/167) before this one.

clrux:
@talbs Okay, rebased with fresh master. Take another look, please.

talbs:
Mind squashing your commits? Once that's done, :+1:

PR #167, opened 2015-09-14
Updating Gulp JS compile for PL JS

clrux:
This small piece of work fixes JS-related Gulp tasks by correcting the source and destination paths for JS files in our `pattern-library/js` folder. Additionally, it corrects a path in `ui.js` that pulls in the correct file.

- [x] @talbs


gulp/tasks/pldoc_scripts.js
```diff
@@ -17,3 +17,14 @@ gulp.task('pldoc_scripts', ['pldoc_scripts-lint'], function() {
...
+gulp.task('pldoc_scripts', ['pldoc_scripts-lint'], function() {
+    return gulp.src([
+        // setup sequence for files that are loaded later with RequireJS
+        './pattern-library/js/*'
+    ])
+    .pipe(uglify())
+    .pipe(gulp.dest(config.pldoc_local)) // move for just browsercync + local preview
+    .pipe(gulp.dest(config.pldoc_dest))
+    .pipe(browserSync.reload({ stream: true }))
+});
```
clrux: @talbs I added a separate rule here to move any JS files that are later loaded indirectly with RequireJS (for example, via ui.js). We may want to use a variable for the path, but for some reason this makes my head hurt.

pldoc/static/js/ui.js
```diff
@@ -3,7 +3,7 @@ define([
     '/public/pldoc/js/tabs.js',
     '/public/pldoc/js/size-slider.js',
     '/public/pldoc/js/color-contrast.js',
-    '/public/js/select-replace.js'
+    '/public/pldoc/js/select-replace.js'
```
clrux: The pathing here was incorrect. Fixed now.

gulp/tasks/pldoc_scripts.js
```diff
@@ -17,3 +17,14 @@ gulp.task('pldoc_scripts', ['pldoc_scripts-lint'], function() {
         .pipe(gulp.dest(config.pldoc_dest))
         .pipe(browserSync.reload({stream:true}))
 });
+
+gulp.task('pldoc_scripts', ['pldoc_scripts-lint'], function() {
+    return gulp.src([
+        // setup sequence for files that are loaded later with RequireJS
+        './pattern-library/js/*'
```
talbs: Yeah, I would try to make this into a configuration variable. I think ``pldoc_src_files`` makes sense if you want to sync it up with your pathing change or if you want to try to convert/simplify and just make ``pldoc_src`` this value.

gulp/tasks/pldoc_scripts.js
talbs: See [my note above on using a config variable](https://github.com/edx/ux-pattern-library/pull/167/files#r39582587). I think its doable and not too headache inducing.

pldoc/static/js/ui.js
talbs: D'oh!

talbs:
@clrux, thanks for the clean up. I left one comment about trying to clean up and use the settings already contained in https://github.com/edx/ux-pattern-library/blob/clrux/js-gulp-compile-fix/gulp/config.js#L69.

Let me know if that works and I'll take another quick look before giving thumbs.

talbs:
@clrux, also one more thing noted, I'm seeing the following untracked file now after running the ``gulp`` command on your branch locally:

```git
Your branch is up-to-date with 'origin/clrux/js-gulp-compile-fix'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	public/pldoc/js/select-replace.js

nothing added to commit but untracked files present (use "git add" to track)
```

Is this something that should not exist? If not, why is it here?

talbs:
> src was already a defined variable pointing to this location, so I just used it. Cool?

Cool.

> Also, the untracked file should now be tracked. Not sure why it wasn't included in the commit, especially when I used git add .. It's the proper, newly compiled JS that's part of the additional Gulp rule.

Cool Cool.

talbs:
:+1:

talbs:
Oh, please squash your commits too. Thx for the fixes!

PR #169, opened 2015-09-22
FIX - PL Doc: Correcting scss link paths for elements

talbs:
This work fixes and abstracts even more scss link paths per element in the documentation site. This got out of whack with the recent directory changes needed to make the pattern library a versioned Bower Package.

 - - -

##### Reviewers

* [x] FED/Doc Site MGMT - @frrrances


frrrances:
Nice clean cleanup! :moneybag: :truck: :+1:

PR #170, opened 2015-09-25
Icon font work

clrux:
This is a work in progress.

## What I'm trying to do...

Using ["The new Kellum method"](http://scottkellum.com/2013/10/25/the-new-kellum-method.html), which, at least from initial impressions, meets our requirements, I'm trying to implement this method into our library using packages we're already using, such as our `edx-svg` font library.

## How does this shiz work?

Icomoon spits out actual fonts in addition to the SVG definitions file we've been using. We basically create a second font (i.e., `iconfnt` in this case). It's basic font replacement, except we're [replacing glyphs/words](http://www.w3.org/TR/SVG/fonts.html#GlyphElementUnicodeAttribute) with SVG paths, rather than a TTF or WOFF (like with Open Sans).

If the user specifies a font that overrides our Open Sans or Iconfon fonts, the glyphs fall back to the word. There are no funky ISO boxes or shapes to confuse the user or screen reader. There's either the natural language word we're replacing or nothing.

The example I'm working with is the `video-camera` icon - the first in the set.

## Why isn't this shiz working?

Not sure :stuck_out_tongue_closed_eyes: 

## Notes

The Gulp workflow jacked up the file paths, so I had to remove a bunch of extraneous things that were duplicated, in the wrong spot, and not being used. That's what all the deletions are.

cc @talbs @cptvitamin


talbs:
Should this be closed given that we're actively working on https://github.com/edx/ux-pattern-library/pull/173?

PR #171, opened 2015-09-28
PL Doc: Add in Additional Asset Tasks to Gulp Workflow

talbs:
This work adds in quick support for moving both icons and fonts to the proper UXPL Documentation Site paths needed for both production and for local development.

**NOTE**: This work only moves assets using the default gulp task (running ``gulp``). It does not watch assets yet. A follow-on PR will tackle that.

- - -

#### Reviewers

* [ ] Directory structure/Gulp workflow - @clrux

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -51,13 +51,13 @@ module.exports = {
     images:                     {
         // pattern library
         src:                    './pattern-library/images',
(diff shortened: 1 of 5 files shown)


clrux:
:+1: Thanks!

PR #173, opened 2015-09-29
New font icons, Modernizr, and A Font Garde

clrux:
## Accessible icon font and fallbacks

Our pattern library previously used SVG paths for icons. While this method was accessible, it required a bunch of markup and the SVG paths were difficult to scale alongside text using font size. Going back even further, we used CSS-based icon fonts that relied on the `:before` pseudo attribute, and while incredibly easy to use, broke if `@font-face` wasn't available or if the font icons didn't load. Users would end up with unhelpful glyph boxes. This was especially painful in instances where the icon was the only indication of the action.

### A Font Garde + Modernizr

The approach we're using here comes from The Filament Group's blog post ["Bulletproof Accessible Icon Fonts"](https://www.filamentgroup.com/lab/bulletproof_icon_fonts.html) and uses feature detection plus a nifty piece of JavaScript to check if the font icons load. If they do, a class is appended to the `html` of the document and the icon fonts render using that class. If the test fails, fallbacks are displayed instead.

#### Simplified markup

Rather than referencing an SVG and its definitions file, we now use one of the following fallback options:

##### Critical icons that fallback to text
```
<span class="icon-fallback-text">
    <span class="icon icon-video-camera" aria-hidden="true"></span>
    <span class="text">Video</span>
</span>
```

##### Critical icons that fallback to an image
```
<span class="icon-fallback-img">
    <span class="icon icon-video-camera" aria-hidden="true"></span>
    <span class="text">Video</span>
</span>
```

#### When font icons are available

When the tests pass, the name of the icon font (`.edx-icons` in our case) is appended to the `html` of the document, alongside a couple of Modernizr check classnames that are used by the `afontgarde.js` file indirectly.

We then style our icon fonts to rely on this class if it's present:

```
.edx-icons .icon {
    font-family: $icon-font;
    ...
}

.edx-icons .icon-linkedin-square:before {
    content: "\e600";
}
...
```

#### When font icons are _not_ available

If the font icon doesn't load, if `@font-face` isn't available in the browser, or if the user chooses not to let the site use its available stylesheets, we don't want random, unhelpful glyph boxes. Since our icon fonts look for the appended `html` class they won't load, giving us a chance to create fallbacks.

##### Critical icons that fallback to text

Just the text in `text` is displayed. The icon itself is hidden.

##### Critical icons that fallback to an image

We create separate rules to use CSS background images instead, like the following:

```
.icon-linkedin-square {
    background: url('path-to-image') no-repeat;
}
```

## Notable changes

[...]

PR #174, opened 2015-09-29
Correcting "New font icons, Modernizr, and A Font Garde"

talbs:
This work revises and builds upon the WIP PR https://github.com/edx/ux-pattern-library/pull/173 to make the conversion to the new icon method near complete.

- - -

##### TODOs

* [ ] Fix the rest of the script linting warnings found with the new ``modernizr.fontface-generatedcontent.js`` vendor file
* [ ] along side https://github.com/edx/ux-pattern-library/commit/7662981f6c2288aa3b8875b4cf86a7a686390124, address revising the syntax of icons used in enhanced form select/dropdown fields.


talbs:
@clrux, please review and let me know when this can be merged into your work to get this very close to done. Also, see the TODOs for once it is merged. I'll need you to tackle the rest of those as they're JS-based with some things you've written.

FYI, @frrrances

talbs:
@clrux, also note all of my commits with the ``fixup!`` prefix. Once merged, please try to squash them into any commits that are already starting related work or rename them to be standalone commits for this repo's history.

pattern-library/sass/_edx-pattern-library.scss
```diff
@@ -21,6 +21,7 @@
 @import 'utilities/normalize';
 @import 'utilities/reset';
 @import 'utilities/typography';
+@import 'utilities/afontgarde';
```
clrux: Does this work for you given it's a CSS file?

pattern-library/sass/utilities/_afontgarde.scss
```diff
@@ -0,0 +1,57 @@
+/*! afontgarde - v0.1.6 - 2015-03-13
```
clrux: Nevermind, you made it a Sass partial. Kewl.

clrux:
It looks like you've taken care of nearly all of the PR feedback from my PR, so thanks. This looks good.

:+1:

pattern-library/sass/utilities/_afontgarde.scss
talbs: Yeah, I did the same with ``_syntax.scss`` borrowed code used for code syntax styling in the Doc Site. Its not worth keeping things as ``.css`` files in a static vendor dir if they're not a bower package IMHO. I'll note in the README that A Font Garde is another tool we're using.

talbs: I commented in a known Sass github issue around being able to import plain old CSS files in a Sass import (https://github.com/sass/sass/issues/279#issuecomment-136926782), FYI.

talbs:
Thanks much for the response, @clrux. I'll merge and you can wrap up loose ends while @frrrances and @cptvitamin have a once over to be cool with firming up this really foundational work.

clrux:
@talbs I've fixed the second of the to-dos, but... with the first (regarding Modernizr), I'm hesitant to fix the warnings for two reasons:

* it works as it is
* it's stock as it is

Annoying as they may be... what do you think, sir?

talbs:
@clrux, thanks for the consideration and mindfulness. I'd still lean on making these syntax fixes as:

* We're already not treating this as a full blown vendor package (not using bower or a vendor dir)
* We're less likely to have to update this solution in the future than other libraries
* This version of modernizr is already customized by Filament Group
* If we don't correct these issues, anyone using the UXPL gulp workflow will be presented with linting errors/warnings every time they compile assets (and this could happen in other places using the UXPL bower package that also use linting on their scripts).
* I've already "beautified" the vendor JS to expand it and have made the syntax changes I could to reduce lint warnings

**With that mind, can you fix the syntax so that no lint warnings appear during the ``gulp`` command?**

If you still feel queasy about it (which I can understand), the real right thing to do is to 1) fix the syntax so no warnings appear, 2) re-uglify the file, and 3) make a PR against A Font Garde to get these syntax fixes into their repo. I'm all for that and even if you don't want to or I can't, I probably will.

PR #176, opened 2015-10-01
PL Doc: Abstract Production/Development Build-centric Gulp Tasks

talbs:
This work takes the good recommendation @frrrances had, and the practice that I've personally set up at https://github.com/talbs/mise-en-place/tree/master/gulp/tasks, and creates ``build-production`` and ``build-development`` Gulp tasks. the ``default`` task has been simplified and now uses the ``build-development`` task.

Additionally, the ``styles`` task and config settings have been revised to:

* not contain any production-centric steps (minifying CSS now happens within ``build-production``, which should be run before committing any compiled assets.)
* remove any production-centric settings that were not being used by the new tasks.

- - -

##### Reviwers

* [x] @frrrances - Gulp logic/abstraction & task use
* [ ] @AlasdairSwan - Gulp/CSS workflow

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -22,11 +22,6 @@ module.exports = {
             sourcemapsLocation: '.'
         },
 
-        settings_production: {
-            outputStyle: 'compressed',
-            autoprefixer: { browsers: ['last 2 version'] }
-        },
-
         // pattern library
         src_files:              src + '/sass/**/*.scss',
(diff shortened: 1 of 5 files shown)


talbs:
@frrrances, this has been rebased off of master to catch all of the big icon changes, FYI.

frrrances:
:+1:

PR #178, opened 2015-10-02
Update Icon System to be Font-based with a11y fallbacks (via A Font Garde)

clrux:
_Note, this replaces #173 as I was having issues squashing commits. This also gave me an opportunity to make the branch name a little more accurate. You can find commentary and processes in the closed PR, should you need it._

## Accessible icon font and fallbacks

Our pattern library previously used SVG paths for icons. While this method was accessible, it required a bunch of markup and the SVG paths were difficult to scale alongside text using font size. Going back even further, we used CSS-based icon fonts that relied on the `:before` pseudo attribute, and while incredibly easy to use, broke if `@font-face` wasn't available or if the font icons didn't load. Users would end up with unhelpful glyph boxes. This was especially painful in instances where the icon was the only indication of the action.

### A Font Garde + Modernizr

The approach we're using here comes from The Filament Group's blog post ["Bulletproof Accessible Icon Fonts"](https://www.filamentgroup.com/lab/bulletproof_icon_fonts.html) and uses feature detection plus a nifty piece of JavaScript to check if the font icons load. If they do, a class is appended to the `html` of the document and the icon fonts render using that class. If the test fails, fallbacks are displayed instead.

#### Simplified markup

Rather than referencing an SVG and its definitions file, we now use one of the following fallback options:

##### Critical icons that fallback to text
```
<span class="icon-fallback-text">
    <span class="icon icon-video-camera" aria-hidden="true"></span>
    <span class="text">Video</span>
</span>
```

##### Critical icons that fallback to an image
```
<span class="icon-fallback-img">
    <span class="icon icon-video-camera" aria-hidden="true"></span>
    <span class="text">Video</span>
</span>
```

#### When font icons are available

When the tests pass, the name of the icon font (`.edx-icons` in our case) is appended to the `html` of the document, alongside a couple of Modernizr check classnames that are used by the `afontgarde.js` file indirectly.

We then style our icon fonts to rely on this class if it's present:

```
.edx-icons .icon {
    font-family: $icon-font;
    ...
}

.edx-icons .icon-linkedin-square:before {
    content: "\e600";
}
...
```

#### When font icons are _not_ available

If the font icon doesn't load, if `@font-face` isn't available in the browser, or if the user chooses not to let the site use its available stylesheets, we don't want random, unhelpful glyph boxes. Since our icon fonts look for the appended `html` class they won't load, giving us a chance to create fallbacks.

##### Critical icons that fallback to text

Just the text in `text` is displayed. The icon itself is hidden.

##### Critical icons that fallback to an image

We create separate rules to use CSS background images instead, like the following:

```
.icon-linkedin-square {
    background: url('path-to-image') no-repeat;
}
```

## Notable changes

There were a number of changes in this update. A summary of them is as follows:

* removed the SVG icons and their definitions file
* added a new icon font set (generated with the same settings file at Icomoon)
* added a new `afontgarde` package to Bower
* added a new `modernizr` package to Bower
* added both of the above to the compile and load
* updated all of the example icons
* updated the sizing script to update font size

### Post-Icomoon required changes

While we are able to utilize _most_ of the stock icons from Icomoon, we do need to make a few easy changes in order for this process to work.

* Icomoon spits out a `.css` file which I copied and placed into the `_icons.scss` partial.
* `.edx-icons` will need to be added in front of every `.icon` in that same partial
* place any fallbacks at the bottom

## Browser testing

This work was checked in the following browsers:

| Platform | Browser | Version | Passed |
|----------|---------|---------|--------|
| Mac      | Chrome  | 45      | Yes    |
|          | Safari  | 8       | Yes    |
|          | Firefox | 40      | Yes *  |
| Windows  | Chrome  | 45      | Yes |
|          | Firefox | 40      | Yes * |
|          | IE      | 10      | Yes  |
|          | IE      | 11      | Yes  |

_* A note about Firefox: Firefox has a known bug where it doesn't load or download local fonts. So, previewing the fonts with `@font-face` disabled will result in the fallbacks being triggered. You'll need to view this from a server to see it in action.

## Documentation

View our [Icons styleguide](https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Icons) for documentation and use.

## Reviewers

- [x] @talbs (Sass and overall overallness)
- [x] @frrrances  (second comprehender)


clrux:
@talbs Do you want everything squashed into a single commit?

talbs:
@clrux, per your squashing question, I'm fine if this all gets squashed down to one commit, but please provide some descriptions on new lines in the commit, like so:

```
Revising icon font rendering method (using A Font Garde library)


* pldoc: adjusting icon element preview/rendering
* pldoc: moving show 'styling' tab logic into Jekyll, not Sass
* pldoc: pdating size slider
* update README with A Font Garde reference in third party libs
```

pldoc/static/js/pattern-library.js
```diff
@@ -3,6 +3,9 @@ require.config({
     paths: {
         jquery: "/public/js/jquery.min",
         svg4everybody: "/public/js/svg4everybody.min",
+        modernizr: "/public/js/modernizr-custom.js",
+        afontgarde: "/public/js/afontgarde.js",
+        edxicons: "/public/js/edx-icons.js"
     },
     shim: {
         'jquery': {
```
clrux: Proposal from @andy-armstrong, add this to the shim after 'jquery':

    'afontgarde': {
        export: 'AFontGarde'
    }

You can then change the initialize.js to be:

    define([
    'afontgarde' // require this for this script to work
    ], function(AFontGarde) {
    AFontGarde('edx-icons', {
        // Check a few random icons to see if our font loads
        glyphs: '\uE621\uE622\uE623'
    });

clrux:
@talbs I've squashed this into a single commit and updated the commit message per your recommendation. Just waiting on @frrrances' feedback now. Let me know if you guys want me to walk you through it.

talbs:
@clrux, thanks. A few small NITs/questions:

Your squashed commit message should read like a step in time, this includes the listed details in the later lines. With that in mind, please remove the ''PR feedback" and "PR feedback from Andy" lines, as we can assume that any feature/improvement has gone through the review process as part of getting merged. No need to call it out for all time in the repo's history

Also, your squashed commit's main message seems a bit off. You didn't add new icons, you changed the way we reference and render them, i.e. using an Icon font technique with A Font Garde to help for a11y/fallback support. Please change this main message to be more clear for the history of the repo.

If there are any steps in the main body of work to walkthrough for what you did, please note them as bullets at the top of the list in the detailed commit message lines. It seems light and most lines are coming from maintenance commits/steps I did from your branch.

In short, I want this commit message to be communicate what this work did clearly to anyone who's trying to follow along. Since this is a big change, I want to avoid any confusion around what was done.

_posts/elements/2015-04-15-icons.md
```diff
@@ -1,29 +1,33 @@
...
 - pattern
 - icons
 
 slug:       icons
+scssurl:    "components/_icons.scss"
 
 doc:        Styleguide:-Icons
 
-description:  A collection of accessible SVG icons available for using within the edX platform.
+description:  A collection of accessible icons available for using within the edX platform.
```
talbs: NIT: this should mention that these icons are font-based and I think we should remove the "available for using within the edX platform", since the PL is not in the platform currently. "available for use when building UI with the edX Pattern Library"

talbs:
@clrux, why was ``public/images/bg-overlay-example.png`` deleted and re-added as part of this branch's work? I'm not seeing how this is related to your icon adjustments.

clrux:
Not sure, didn't touch it.

talbs:
> Not sure, didn't touch it.

@clrux, okay. I'll see what I can do about it.

clrux:
Did the file itself change? If not, is it worth worrying about?

talbs:
> Did the file itself change? If not, is it worth worrying about?

It looks like not and even with a copy of master's version of the file into this branch there were no diffs. It doesn't look harmful but I don't like having files that weren't touched as part of the diff.

clrux:
Huh, yeah, weird dude. I'm not sure :(

talbs:
@frrrances has signed off on this direction!

PR #191, opened 2015-10-14
FIX: Correcting Broken ToC links in README

talbs:
This corrects some updated heading and skip links in the README.

- - -

##### Reviewers

* [x] @dsjen


dsjen:
:+1:

PR #187, opened 2015-10-14
WIP -- Hackthon: Adding the UI Toolkit

talbs:
WIP between @dsjen and @talbs.

PR #195, opened 2015-10-15
Alasdair/add dropdown menu view

AlasdairSwan:
@talbs I rebased but am seeing 2 of your commits. Can you please take a look at my code and see what you think? I resolved the template issue I was having.
cc @andy-armstrong @dsjen 

Note: after running ``` gulp production ``` all of the font files are deleted and there are a ton of public/js changes. Thus I have not added these changes in here until we have decided upon a better strategy going forward (e.g. adding public/js as a single rjs bundle)


talbs:
@AlasdairSwan, is this work still active? I've closed https://github.com/edx/ux-pattern-library/pull/187 and https://github.com/edx/ux-pattern-library/pull/217 since they weren't ready for merging and aren't being worked on/rebased currently.

I think this work you've done is based on/dependent on some of those closed bits, correct?

AlasdairSwan:
@talbs let me take another look at this today and I will get back to you re. whether this is just junk or if I can salvage it.

PR #197, opened 2015-10-15
PL Doc: Set current nav item to is-current on page load.

jamiefolsom:
Hi @talbs and @AlasdairSwan,

Another super small enhancement. Feel free to use or not. 

**Problem**
When loading a page from permalink, the current nav item is not set.

**Solution**
Compare the window.location.pathname to each nav item in succession, until we reach the one which matches. Then return false to stop the loop.

**How to test**
Check out this branch, navigate to a permalink, and you will note that the current nav item is highlighted.

Thanks!

Jamie


pldoc/static/js/ui.js
```diff
@@ -20,6 +20,12 @@ define([
 
         setupHtml: function() {
             $('html').removeClass('no-js');
+            $("#pldoc-nav-elements ul li a").each(function (index, element) {
```
talbs: NIT: can we keep this selector to class-based attributes only, e.g. ``#pldoc-nav-elements .nav-list-item .pldoc-link``?

talbs:
@jamiefolsom, I had one small inline NIT, but this is awesome and looks good to me. I'll leave @AlasdairSwan to review your JS syntax a bit more in detail, but you have my :+1:

pldoc/static/js/ui.js
```diff
@@ -20,6 +20,12 @@ define([
...
+                if ( $( element ).attr( "href" ) === window.location.pathname ) {
+                  $( element ).addClass("is-current");
+                  return false;
+                }
+            });
```
AlasdairSwan: NIT: can you please update to use single quotes instead of double for consistency.

talbs:
@jamiefolsom, two other things: 

* OCD-wise, would you prefixing your commit message with ``pldoc:``. This helps clarify when we're publishing the doc site to ux.edx.org and also bumping up bower package versions for any UXPL pattern work what commits and work affected what.
* Please add yourself to the [AUTHORS list](https://github.com/edx/ux-pattern-library/blob/master/AUTHORS) :smile:

pldoc/static/js/ui.js
```diff
@@ -20,6 +20,12 @@ define([
 
         setupHtml: function() {
             $('html').removeClass('no-js');
+            $("#pldoc-nav-elements ul li a").each(function (index, element) {
+                if ( $( element ).attr( "href" ) === window.location.pathname ) {
+                  $( element ).addClass("is-current");
+                  return false;
```
AlasdairSwan: Why return false?

pldoc/static/js/ui.js
jamiefolsom: No problem, will do.

pldoc/static/js/ui.js
jamiefolsom: No problem, will do!

jamiefolsom:
Hi @talbs, I think I've addressed those items. Thanks!

pldoc/static/js/ui.js
jamiefolsom: That breaks out of the each() when we find the href that matches
window.location.pathname, preventing unnecessary iterations.

pldoc/static/js/ui.js
```diff
@@ -20,6 +20,12 @@ define([
 
         setupHtml: function() {
             $('html').removeClass('no-js');
+            $('#pldoc-nav-elements .nav-list-item .pldoc-link').each(function (index, element) {
+                if ( $( element ).attr('href') === window.location.pathname ) {
+                  $( element ).addClass('is-current');
```
AlasdairSwan: One last nit: it would be more performant to create a ``` var $el = $( element ); ``` at the start of each loop instead of creating the same jQuery element object twice.

AlasdairSwan:
Once the element jQuery nitpick is resolved good to merge :+1:

jamiefolsom:
Good point, will do.

jamiefolsom:
All set with that improvment, @AlasdairSwan. Thanks!

AlasdairSwan:
Looks good to merge @jamiefolsom  :+1:

talbs:
Nice work and thanks for the UI fix, @jamiefolsom. :truck:ing this to Mergetown now.

jamiefolsom:
:+1: thanks!

PR #198, opened 2015-10-15
PL Doc: Correct Artifacts of Old Icon Rendering

talbs:
This work cleans up a few things that were missing at the end of https://github.com/edx/ux-pattern-library/pull/178, namely:

* syncing up the gulp ``production`` task
* updating the CHANGELOG file
* removing svg4everbody (used for the older SVG icon rendering method) as a bower dependencies

- - -

##### Reviewers

* [x] @AlasdairSwan

PR #201, opened 2015-10-16
PL Doc: Ignore Source Maps

talbs:
This work removes css source maps from production and the rep (per @AlasdairSwan's comment in https://github.com/edx/ux-pattern-library/pull/198#issuecomment-148750609).

- - -

##### Reviewers

* [x] @AlasdairSwan


AlasdairSwan:
:+1:

PR #202, opened 2015-10-16
PL Doc: Use the term "pattern" rather than "element"

talbs:
This work switches tousing the term 'pattern' rather than 'element' for clarity and simplicity throughout docs.

- - -

##### Reviewers

* [x] Overall - @frrrances


frrrances:
look at you with the words! :+1:

PR #204, opened 2015-10-18
Correct, Optimize, and Re-organize edx-icons Fallback Images

talbs:
This work focuses on improving our icon solution by tackling the following:

* corrects a mistake where fallback edx-icons SVG images were not included in the pattern-library source code during the work of https://github.com/edx/ux-pattern-library/pull/178
* optimizes all current edx-icons SVG image files with respect to performance/file size
* partially automates fallback styling rules used with edx-icons image fallbacks

- - -

##### Reviewers

* [ ] FED and repo organization - @frrrances


talbs:
Since the diff on this PR is [huuuuuugee](http://big.assets.huffingtonpost.com/KknQODNpux.gif), here's a rundown of where files went:

* I removed all of the SVG images that serve as fallbacks to our icons from where they were incorrectly placed (`ux-pattern-library/pldoc/static/images/icons`).
* I then moved them into the pattern-lirbary package's source code (alongside other edx-icons assets) at `ux-pattern-library/pattern-library/fonts/edx-icons/fallback-img`
* the above files were then optimized using our `gulp-images` task/workflow
* lastly, I added some additional scaffolding (configuration and an automated each Sass loop) into ``_icons.scss``, which sets the background-image rules for all fallback-img cases devs set in their markup.

pattern-library/sass/patterns/_icons.scss
```diff
@@ -6,6 +6,119 @@
 // NOTE: Read more at https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Icons
 
 // ----------------------------
+// #CONFIG
+// ----------------------------
+$edx-icons: linkedin-square
```
frrrances: This is nice abstraction, but two nits: 
- Can you put "linkedin-square" on it's own line like the rest?
- Is there anyway to grab these from the source so we don't have to update it when a new one is added? Or is this our custom list?

frrrances:
[Huuuuge diff](http://stream1.gifsoup.com/view5/2275083/huge-wave-o.gif) but solid work. :+1:

talbs:
@frrrances, thanks for the notes. I moved that first icon def to a new line like you said. Merging with the other note to look into ways to automate the shared list of support icons between our settings and scss. Thanks!

PR #205, opened 2015-10-19
PL Doc: Jekyll Architecture Tidying

talbs:
This work takes the opportunity to address the following with the Jekyll templating architecture that powers the UXPL Doc Site:

* abstracts out 1 single pattern template to be shared across both index/listing views and detailed pattern views (mad props to [this stackoverflow thread](http://stackoverflow.com/questions/26742959/how-can-i-pass-variables-between-dry-templates-and-ncludes-in-jekyll)
* creates an area for individual metadata models using the ``_data/`` directory (see http://jekyllrb.com/docs/datafiles/).
* syncs up naming conventions and spacing for element metadata in pattern/template files
* updates information/corrects typos

- - -

##### Reviewers

* [x] Overall template architecture - @frrrances


frrrances:
squeaky clean! :gem: :+1:

PR #206, opened 2015-10-19
Organizing Sass Compilation for Better Extensibility and Customization

talbs:
To match some of the standards out there with [Bootstrap](https://github.com/twbs/bootstrap-sass#sass), [Foundation](https://github.com/zurb/foundation/blob/master/scss%2Ffoundation.scss), and other UI Toolkits, this work organizes sass compilation for better extensibility and customization. 

##### The general goals are to:

* allow individual patterns to be called in an app's Sass compilation (if so desired through a custom Sass compile)
* allow for better isolation of dependencies individual patterns may have, which should make building new patterns easier to do without updating so many global settings or manually proofing existing patterns

##### Steps include:

* simplify Sass directory structure and terminology
* create and import _core.scss settings/utilities in one partial
* allow patterns to define settings and helpers scoped to their rendering - global settings has been reduced
* Sass terminology and comment formatting clean up

##### Some terminology changes include:

* components -> patterns (see https://github.com/edx/ux-pattern-library/pull/202)
* variables and config -> settings
* utilities -> global

- - -

##### TODOs

* [x] Abstract variables that are specific to patterns into the individual pattern's .scss partial (within the ``#SETTINGS`` area).
* [x] Sync up any pattern settings that should be overridden in the global ``_settings.scss`` partial
* [x] Update README and Wiki Documentation to reflect new paths/terminology as well as new method for how to customize an app's Sass compilation to pull in new specific isolated pattern partials. 

- - - 

##### Reviewers

* [x] @frrrances - FED and Sass best practices
* [x] @AlasdairSwan  - Ease of use/consuming
* [x] @ormsbee - Core


ormsbee:
So while this looks good to me, when it comes to sass/css, I'm kind of like a caveman with a camera. Are there specific questions that I should try to address?

talbs:
@ormsbee and all, here are some concerns I definitely want poked and proded with this work:

* does calling in settings into each pattern's Sass cause bloat in the compiled CSS (hopefully not since its mostly Sass vars, functions, and helpers)?
* can we build new patterns easily using this core and not leave anything out or have to re-invent the wheel with settings per pattern?
* is this way of organizing things more easy to follow for consumers of the UXPL (e.g. @AlasdairSwan and the Programs work)?
* how hard or easy would this change make upgrading UXPL versions for apps/installations?

talbs:
Also, rebasing this against ``master`` a bit later today.

talbs:
Freshly rebased! Have at it at your leisure, amigos :)

talbs:
STATUS: I've chatted with @frrrances about the approach of this abstraction work. We found the following valuable:

* a foundational Sass partial that manages including any global settings, functions/helpers, and foundational resets (``_normalize.scss`` and ``_reset.scss``) - aka ``global/_core.scss``
* settings that are scoped to specific patterns should be defined within the pattern's individual Sass partial

The one directional change was to not include core in each pattern's Sass partial (as some of the reset-based code there will cause redundant lines of CSS), but rather to call the ``global/_core.scss`` partial before any pattern calls (either the default all released patterns contained in https://github.com/edx/ux-pattern-library/blob/talbs/org-core%2Bpatterns/pattern-library/sass/_edx-pattern-library.scss or in a customized list called by the app's styling).

- - -

I've updated this branch to address that work and tackled all of the TODOs noted above as well - this means that documentation (README and Wiki Docs) are up to date.

- - -

@frrrances, this is ready for your primary review. 

@AlasdairSwan, both Frances and I agreed that the scoped way you're using the UXPL isn't one of the primary ways we are expecting this library to be consumed. Happy to support your customization work for your needs, but I don't think we want to architect any scoping-based variables as part of this PR. I'd still like your eyes on this work and your review as well.

@ormsbee, happy to give you an architectural walkthrough if things are still unclear.

README.md
```diff
@@ -349,24 +379,20 @@ the ticket to indicate that you are working on it. Don't hesitate to ask
...
 
 #### Submitting Code
-
 For code contributions, please open up a pull request! PRs will get OSPR tickets
 assigned to them, as mentioned in the above contributing guidelines.
 
 #### Approval by UX and Front End Team Members
-
 An edX UX Team member will be working with you on any pull requests you make.
 They be evaulating your pull request from a design point of view as well as from
```
frrrances: Just noticed that this is missing a "will" - "They will be evaluating..." (Unless we really are that hip.)

frrrances:
@talbs Thanks for the good slicing. I am happy with this approach and think it tackles both extensibility and performance for the main PL use cases. :+1:

AlasdairSwan:
Looks good to me @talbs. I like move the partial specific settings into the partial itself. It will make it much easier to use said partials in isolation. :+1:

ormsbee:
:+1: :)

README.md
talbs: Thanks for the good find. Fixed that right up in https://github.com/edx/ux-pattern-library/pull/229.

talbs:
Thanks, all!

PR #207, opened 2015-10-19
Removing Colors Sass Partial from Official Pattern Styling Directory

talbs:
This work removes ``_colors.scss`` as an official sass pattern partial since it was only providing supportive styling for the documentation site. The contents of that file are now placed within the documentation site's Sass

- - -

##### Reviwers

* [x] Overall FED and Sass Architecture - @frrrances


pldoc/static/sass/_components.scss
```diff
@@ -498,9 +723,49 @@
...
+// ui depth
+.pldoc-element-depth {
+
+    // padding for demo purposes
+    .depth {
+        padding: spacing-vertical(small) spacing-horizontal(base);
+    }
+
+    // light text on darker backgrounds for demo purposes
+    .depth--2,
```
frrrances: double dash! does that just add a little bling or does it have meaning?

pldoc/static/sass/_components.scss
```diff
@@ -342,6 +364,209 @@
 // color swatches
 .pldoc-pattern-colors {
 
+    .brand-primary {
```
frrrances: How come the classes? Would you be adding these as classes on the DOM? 

```
<p class="plddoc-pattern-colors brand-primary base">Give me a base background color, yo!</p>
```

frrrances:
Good clean up, just a couple small nittish questions... Once you let me know I'm completely missing the point, I'm all good with :truck: :house: :+1:

pldoc/static/sass/_components.scss
talbs: Good question, the classes are just for the actual rendered color palette in http://ux.edx.org/elements/colors/. These are not meant to be used as official classes that the UXPL bower package supports in a proj - hence why I'm moving all of this Sass/CSS to be part of the Documentation Site and not the package.

Seem still legit?

pldoc/static/sass/_components.scss
talbs: Oh, this isn't BEMing or anything, this class (a product of the [ui-depth](https://github.com/edx/ux-pattern-library/blob/master/pattern-library%2Fsass%2Fpatterns%2F_depth.scss#L92) pattern) is just signifying a UI element that has a depth of negative 2. If I could, I'd write it as ``.depth-(-2)``, but that's not valid CSS.

Make sense?

frrrances: haha! makes perfect sense now that i look at it in context. carry on!

PR #222, opened 2015-11-02
Simplify font-family Styling + Inheritance

talbs:
Per the good issue brought up in https://github.com/edx/web-certificates/pull/18#discussion_r43634913, a project that relies on the UXPL, this work simplifies font-family styling to rely on global reset vs. specific heading/copy/form definitions.

The goal is to allow for easier typography customizations/overrides without having to:

1. litter individual rules within an app with resets/changes, leading to larger file size and messier maintenance.
2. add cascade/specificity to an app's CSS

- - -

##### Please Note

* The ``$text-base-font-family`` and ``$headings-base-font-family`` variables are still in tact and can be customized. 
* ``$headings-base-font-family`` is not currently used in any headings styling, since they are by default already inheriting their ``font-family`` from the reset. This is merely here to allow for deeper manual customization

- - -

##### Reviewers

* [x] Overall - @frrrances


frrrances:
Man, you are fast! Looks good though. :bullettrain_front: :sparkles: :+1:

PR #227, opened 2015-11-11
Use Sass to Loop and Generate Class-based Rules

talbs:
This work converts some of the class-based rules in the UXPL to:

* be set using simple configuration within each pattern's partial
* be generated using Sass's ``@each`` method

The goal is to reduce redundant code and to help manage sizes/parameters in a single place.

- - -

##### Reviewers

* [ ] FED - @frrrances


public/css/main-rtl.css
```diff
@@ -5208,27 +5208,27 @@ Note: sure .FONT_NAME comes first for adjoining classes bug in IE7.
   }
 }
 
-.depth--3, .pldoc-wrapper-content-supplemental {
+.btn--3, .pldoc-wrapper-content-supplemental {
   background: #282c2e;
 }
 
-.depth--2 {
+.btn--2 {
```
frrrances: did you mean for this to change from depth to btn (here and below)?

public/css/main-ltr.css
```diff
@@ -5208,27 +5208,27 @@ Note: sure .FONT_NAME comes first for adjoining classes bug in IE7.
   }
 }
 
-.depth--3, .pldoc-wrapper-content-supplemental {
+.btn--3, .pldoc-wrapper-content-supplemental {
```
frrrances: did you mean for this to change from depth to btn (here and below)?

pattern-library/sass/patterns/_depth.scss
```diff
@@ -83,30 +91,10 @@ $depth-2-bg:    palette(grayscale, white)   !default;  // +2 depth level
...
-}
-
-.depth-1 {
-    @extend %depth-1;
-}
-
-.depth-2 {
-    @extend %depth-2;
+@each $depth-level in $depth-levels {
+    .btn-#{$depth-level} {
```
frrrances: I think this is the problem, now that I look at it... .btn -> .depth

frrrances:
@talbs This makes it easy to maintain lots of canned classes but we loose a lot of readability, especially around the canned classes that makes it hard to use. I don't expect we'll be adding a ton more canned classes (like we have to do with images), so I'm not sure how much gain we're getting here. Thoughts?

pattern-library/sass/patterns/_depth.scss
talbs: Yep, thanks for the good catch. Copying and pasting got away from me.

talbs:
Gonna close this for now as I think we have larger things to worry about in and out of the UXPL. We can revisit if canned application styling becomes a pain.

PR #230, opened 2015-11-16
Publish: 11/16/2015

talbs:
This set of things to publish to ux.edx.org includes: 
- icons revamp corrections/missing steps
- lots of PL Doc workflow and Jekyll improvements
- Pre-compiled CSS versions (beta) of UXPL base Sass
- improving local development tool maintenance
- highlight current pattern in navigation when on a pattern page
- consistently using "pattern" across templates/copy
- image optimization of icon assets
- better management of bi-app-sass dependency
- organizing sass compilation for better extensibility and customization

PR #231, opened 2015-11-20
Increase placeholder image sizes

rlucioni:
Drive-by contribution. I noticed the current placeholder images don't look so hot on high-resolution displays. This PR includes a commit bumping the sizes of the placeholder images. For example, here's before:

And here's after:

@talbs, impressions? How do you deploy to http://ux.edx.org/?

### DIFF
--- a/_posts/patterns/2015-07-01-grid.md
+++ b/_posts/patterns/2015-07-01-grid.md
@@ -485,94 +485,94 @@ info:               A responsive, mobile first fluid grid system, based on Susy 
     <div class="example-grid example-grid-gallery-semantic">
         <section class="photo-gallery photo-gallery-large">
             <figure class="photo">
-                <img class="photo-source" src="http://placehold.it/200x100" alt="Puppies playing in a field">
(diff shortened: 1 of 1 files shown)


talbs:
Thanks for the drive-by! Awesome to see you in here, @rlucioni. This makes sense as we were probably stretching placeholder's original dimensions on larger view ports (and on retina displays). 

:+1:

talbs:
Also, we deploy with a separate PR off of master to gh-pages (Jekyll and GH hosting take care of the rest). Here's an example - https://github.com/edx/ux-pattern-library/pull/230

I'll open one of those for you once this is merged.

rlucioni:
@talbs sounds good, merging!

talbs:
D'oh. @rlucioni, you should also add yourself to the AUTHORS file (https://github.com/edx/ux-pattern-library/blob/master/AUTHORS) in a follow-up PR. :)

rlucioni:
Done in https://github.com/edx/ux-pattern-library/pull/232.

PR #245, opened 2015-12-28
FIX - Using an Example of a Link as a Button

talbs:
This work aims to adjust the instructions and information around the button pattern's visual styling to help users of the UXPL choose the appropriate semantic element/markup for their needs (regardless of styling).

- - -

##### Reviewers

* [ ] @cptvitamin


cptvitamin:
:+1: thanks for making that guidance more explicit.  Together we will end the use of ```href="#"``` across the entire universe.

PR #249, opened 2016-01-02
Adding Information-based System Color

talbs:
This work adds a system utility-based information color (and range)

* to be used on educational, informational, and note-based UI elements
* to be based off of ``$branding-primary`` setting

- - -

##### Reviewers

* [x] Overall - @clrux


clrux:
:+1: 

Just be aware of the three `error-trans[]` colors with dark text in actual use-cases.

PR #258, opened 2016-01-13
Patterns: Lists - Take 2

talbs:
https://openedx.atlassian.net/browse/UX-2777.

This work contains styling for basic list renderings as a pattern in addition to a few small fixes/abstractions needed along the way, namely:

* correcting pattern styling source paths
* updating settings spacing maps to use vert/horiz baselines
* bulletproofing text/base link styling - links now do not have a bottom visual border by default, but text-based links do.

- - - 

##### What this work does not cover:

* navigation elements
* complex listings (like cards/search results)

- - -

##### Reviewers

* [x] FED, Visual, and UI Design - @frrrances 
* [x] a11y (mechanical and theoretical) - @cptvitamin 

- - -

##### Open questions

* [x] Should we abstract things out to Sass placeholder extends (e.g. ``%list-ordered``)? **Yes**
* [x] Should we consider responsive with horizontal lists? **No**
* [x] How much to class vs. rely on DOM structure (``.item`` vs. ``li``) **Account for basic use of both**
* [x] Do we need nested examples and support? **Yes**


cptvitamin:
:+1: put this in your list of "done"

PR #262, opened 2016-01-20
HTML Boilerplates - For Kickstarting using the UXPL

talbs:
**This work adds in some basic example HTML boilerplates with the purpose of:**

* Communicating how to apply the HTML Styleguide (https://github.com/edx/ux-pattern-library/wiki/Styleguide:-HTML) into base cases
* Giving developers and template-creators a starting point for how to organize their DOM and what basic properties, settings, and elements to start with
* Showing different options for different configuration and base UI rendering support (e.g. favicon, touch icon, IE/older browser warning). These should be removed/used on a case-by-case basis.

**What this work does not do:**
* Create production-ready templates for apps - these still need to be adjusted to fit the flow/content/settings for each application
* Point to production ready assets (CSS, JS) - links and references need to be replaced with each app's real URLs/URIs

- - -

##### Reviewers

* [x] FED and General Knowledge/Use - @dsjen and @andy-armstrong 
* [x] A11y - @cptvitamin

### DIFF
--- a/pattern-library/html/boilerplate-blank-RTL.html
+++ b/pattern-library/html/boilerplate-blank-RTL.html
@@ -1,0 +1,54 @@
+<!DOCTYPE html>
+
+<!--
+edX Pattern Library: Boilerplate HTML Template: Header + Footer (RTL layout direction!)
+
+NOTES:
+* base template with details on how to handle primary/secondary assets,
+  metadata, and basic view-level UI (on canvas vs. conditional and off-canvas elements)
+* right-to-left (RTL) layout configuration have been added
+* see https://github.com/edx/ux-pattern-library/wiki/Styleguide:-HTML#right-to-left-rtl-support
+  for more information
+-->
+<html class="no-js" lang="en" dir="rtl">
+    <head dir="rtl">
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width,initial-scale=1">
+
+        <base href="{base path}">
+
+        <!-- place favicon.ico in the root directory -->
+        <link rel="shortcut icon" href="{favicon path}">
+        <!-- place touch icon in the root directory -->
+        <link rel="apple-touch-icon" href="{icon path}">
+
+        <!-- view/page metadata -->
+        <title>{page/view's title}</title>
+        <meta name="description" content="">
+
+        <!-- place primary assets (needed for or to support base view/page rendering) here -->
+        <link rel="stylesheet" href="{main css path}">
+        <script src="{main js path}"></script>
+    </head>
+
+    <body class="{app/site-section-name} {view-name} {stateful-class}">
+        <!-- if needed, provide a message for older versions of IE -->
+        <!--[if lt IE 9]>
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser.
+            Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view" dir="rtl">
+            <!-- page/view content is placed here -->
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view" id="wrapper-off-view" dir="rtl">
+            <!-- off-canvas navigation and elements -->
+            <!-- injected view-centric alerts are placed here -->
+            <!-- injected modals/dialogues are placed here -->
+        </div>
+    </body>
+</html>
+
--- a/pattern-library/html/boilerplate-blank.html
+++ b/pattern-library/html/boilerplate-blank.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+
+<!--
+edX Pattern Library: Boilerplate HTML Template: Header + Footer
+
+NOTES:
+
+* base template with details on how to handle primary/secondary assets, metadata,
+  and basic view-level UI (on canvas vs. conditional and off-canvas elements)
+-->
+<html class="no-js" lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width,initial-scale=1">
+
+        <base href="{base path}">
+
+        <!-- place favicon.ico in the root directory -->
+        <link rel="shortcut icon" href="{favicon path}">
+        <!-- place touch icon in the root directory -->
+        <link rel="apple-touch-icon" href="{icon path}">
+
+        <!-- view/page metadata -->
+        <title>{page/view's title}</title>
+        <meta name="description" content="">
+
+        <!-- place primary assets (needed for or to support base view/page rendering) here -->
+        <link rel="stylesheet" href="{main css path}">
+        <script src="{main js path}"></script>
+    </head>
+
+    <body class="{app/site-section-name} {view-name} {stateful-class}">
+        <!-- if needed, provide a message for older versions of IE -->
+        <!--[if lt IE 9]>
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser.
+            Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+            <!-- page/view content is placed here -->
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view" id="wrapper-off-view">
+            <!-- off-canvas navigation and elements -->
+            <!-- injected view-centric alerts are placed here -->
+            <!-- injected modals/dialogues are placed here -->
+        </div>
+    </body>
+</html>
--- a/pattern-library/html/boilerplate-header+footer+content+nav.html
+++ b/pattern-library/html/boilerplate-header+footer+content+nav.html
@@ -1,0 +1,122 @@
+<!DOCTYPE html>
+
+<!--
+edX Pattern Library: Boilerplate HTML Template: Header + Footer
+
+NOTES:
+
+* base template with the addition of how to handle an app-centric header/footer semantically.
+* base template with the addition of how to handle different types of navigation (user vs. app-centric).
+* base template with addition of how to start organizing main vs. supplementary/aside content
+* this assumes that the main and aside content are next to each other in the document's flow.
+  This does not necessarily have to be the case, but the same markup could be used in more complex cases.
+-->
+<html class="no-js" lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width,initial-scale=1">
+
+        <base href="{base path}">
+
+        <!-- place favicon.ico in the root directory -->
+        <link rel="shortcut icon" href="{favicon path}">
+        <!-- place touch icon in the root directory -->
+        <link rel="apple-touch-icon" href="{icon path}">
+
+        <!-- view/page metadata -->
+        <title>{page/view's title}</title>
+        <meta name="description" content="">
+
+        <!-- place primary assets (needed for or to support base view/page rendering) here -->
+        <link rel="stylesheet" href="{main css path}">
+        <script src="{main js path}"></script>
+    </head>
+
+    <body class="{app/site-section-name} {view-name} {stateful-class}">
+        <!-- CASE: if needed, provide a message for older versions of IE -->
+        <!--[if lt IE 9]>
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser.
+            Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-header-app">
+                <header class="header-app" aria-label="Page">
+                    <!-- example: app logo/title -->
+                    <h1 class="header-app-title">
+                        <a class="logo" href="{app or company home url}">
+                            <img class="logo-img" src="{path/to/logo-img.svg}" alt="{app/site name} Home Page" />
+                            <!-- example: alternate app/site name (if not a graphical logo) -->
+                            <!-- {app/site name} -->
+                        </a>
+                    </h1>
+                    <!-- insert header content here -->
+
+                    <!--
+                      NOTE: simplified user account navigation example. This would be more complex
+                      in reality (with dropdowns, etc.)
+                    -->
+                    <div class="wrapper-navigation">
+                        <nav class="nav-account" id="nav-account" aria-label="User">
+                            <a href="{url}">
+                                {username}
+                                <span class="sr-only">'s settings/profile</span>
+                            </a>
+                        </nav>
+                    </div>
+                </header>
+            </div>
+
+            <!--
+              NOTE: simplified course navigation example.
+            -->
+            <div class="wrapper-navigation">
+                <nav class="nav-app" id="nav-app" aria-label="Course">
+                    <a href="{url}">Course Home</a>
+                    <a href="{url}">Course Content</a>
+                    <a href="{url}">Discussions</a>
+                    <a href="{url}">Assignments</a>
+                    <a href="{url}">Assets &amp; Utilities</a>
+                    <a href="{url}">My Progress</a>
+                </nav>
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-content">
+                <main class="content-main" id="content-main" aria-label="Content">
+                    <!-- insert main content here -->
+                </main>
+
+                <aside class="content-aside" id="content-aside" aria-label="{aside content/purpose}">
+                    <!-- insert aside/supplemental content here -->
+                </aside>
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" aria-label="Page">
+                    <!-- insert footer content here -->
+                </footer>
+            </div>
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view" id="wrapper-off-view">
+            <!-- off-canvas navigation and elements -->
+            <!-- injected view-centric alerts are placed here -->
+            <!-- injected modals/dialogues are placed here -->
+        </div>
+    </body>
+</html>
--- a/pattern-library/html/boilerplate-header+footer.html
+++ b/pattern-library/html/boilerplate-header+footer.html
@@ -1,0 +1,96 @@
+<!DOCTYPE html>
+
+<!--
+edX Pattern Library: Boilerplate HTML Template: Header + Footer
+
+NOTES:
+* base template with the additon of how to handle an app-centric header and footer semantically.
+* this assumes that the header and footer need to have their own visual band/are full width
+  against the viewport. If that's not needed, the ``wrapper-*`` elements around them can be
+  removed in favor of the ``wrapper-view`` grid/layout/visual control.
+-->
+<html class="no-js" lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width,initial-scale=1">
+
+        <base href="{base path}">
+
+        <!-- place favicon.ico in the root directory -->
+        <link rel="shortcut icon" href="{favicon path}">
+        <!-- place touch icon in the root directory -->
+        <link rel="apple-touch-icon" href="{icon path}">
+
+        <!-- view/page metadata -->
+        <title>{page/view's title}</title>
+        <meta name="description" content="">
+
+        <!-- place primary assets (needed for or to support base view/page rendering) here -->
+        <link rel="stylesheet" href="{main css path}">
+        <script src="{main js path}"></script>
+    </head>
+
+    <body class="{app/site-section-name} {view-name} {stateful-class}">
+        <!-- CASE: if needed, provide a message for older versions of IE -->
+        <!--[if lt IE 9]>
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser.
+            Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-header-app">
+                <header class="header-app" aria-label="Page">
+                    <!-- example: app logo/title -->
+                    <h1 class="header-app-title">
+                        <a class="logo" href="{app or company home url}">
+                            <img class="logo-img" src="{path/to/logo-img.svg}" alt="{app/site name} Home Page" />
+                            <!-- example: alternate app/site name (if not a graphical logo) -->
+                            <!-- {app/site name} -->
+                        </a>
+                    </h1>
+                    <!-- insert header content here -->
+                </header>
+            </div>
+
+            <div class="wrapper-nav">
+                <nav class="nav-app">
+                    <a href="{url}">Home</a>
+                    <a href="{url}">Patterns</a>
+                    <a href="{url}">Styleguides</a>
+                    <a href="{url}">Help &amp; Support</a>
+                </nav>
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-content">
+                <!-- insert content here -->
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" aria-label="Page">
+                    <!-- insert footer content here -->
+                </footer>
+            </div>
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view" id="wrapper-off-view">
+            <!-- off-canvas navigation and elements -->
+            <!-- injected view-centric alerts are placed here -->
+            <!-- injected modals/dialogues are placed here -->
+        </div>
+    </body>
+</html>


talbs:
@andy-armstrong and @dsjen, let me know if this work is valuable enough to review before I depart. I'm happy to wait or try to address anything you guys want before then. Completely your call.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,116 @@
...
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" role="contentinfo">
+                    <!-- insert footer content here -->
+                </footer>
+            </div>
+        </div>
+
+        <!-- if needed, provide off-canvas DOM elements for global alert/message UI -->
+        <div class="wrapper-alerts" aria-hidden="true">
```
andy-armstrong: Is this part of the alert pattern? Should it be if not?

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,89 @@
...
+            </div>
+        </div>
+
+        <!-- if needed, provide off-canvas DOM elements for global alert/message UI -->
+        <div class="wrapper-alerts" aria-hidden="true">
+            <!-- injected view-centric alerts are placed here -->
+        </div>
+
+        <!-- if needed, provide off-canvas DOM elements for dialogue/modal UI -->
+        <div class="wrapper-dialogue" aria-hidden="true">
```
andy-armstrong: Is this standardized? Don't we call the modals now? Also, I think it should be ``wrapper-dialog``. I would be tempted to drop this until we've worked out the modal pattern.

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,89 @@
...
+
+        <base href="/">
+
+        <!-- place favicon.ico in the root directory -->
+        <link rel="shortcut icon" href="favicon.ico">
+        <!-- place touch icon in the root directory -->
+        <link rel="apple-touch-icon" href="apple-touch-icon.png">
+
+        <!-- view/page metadata -->
+        <title></title>
```
andy-armstrong: I like the convention of using {stuff-should-go-here}, so how about making this:

    <title>{page-title}</title>

andy-armstrong:
It might be a good idea to have a right-to-left boilerplate file too.

andy-armstrong:
@dsjen It would be great to find a place to slot these in to your redesigned doc site nav.

andy-armstrong:
@talbs I know that you say that these templates don't point to real CSS/JS, but I think it would be great if these rendered with the appropriate styling on the doc site. Do you think that subverts their intention slightly, or do you think it would be clear enough to folks what to do. I guess we could have a commented section that says that this brings in the UXPL doc site styling, so switch it out for your own.

pattern-library/html/boilerplate-header+footer.html
talbs: Sounds good!

pattern-library/html/boilerplate-header+footer+content+nav.html
talbs: Oh, no. As we chatted about in person, this is meant to be a wrapper or dedicated place to put alert-type of elements in a page's DOM. This was meant to be a guide for how to handle off-canvas or off-view elements that we may want to add to the DOM/UI based on specific events and cases.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,116 @@
...
+            <!-- injected view-centric alerts are placed here -->
+        </div>
+
+        <!-- if needed, provide off-canvas DOM elements for dialogue/modal UI -->
+        <div class="wrapper-dialogue" aria-hidden="true">
```
talbs: @andy-armstrong, heads up that this area was also meant to be a guide for how to handle off-canvas or off-view elements that we may want to add to the DOM/UI based on specific events and cases. In this case its more specifically for dialogues/modals.

pattern-library/html/boilerplate-header+footer.html
talbs: This was meant to be a guide for how to handle off-canvas or off-view elements, specifically modals/dialogues, that we may want to add to the DOM/UI based on specific events and cases.

pattern-library/html/boilerplate-header+footer+content+nav.html
talbs: @andy-armstrong, I'll abstract this area out so its less about specific patterns and is more about location of where to put off-canvas UI elements like alerts, modals, and others.

pattern-library/html/boilerplate-header+footer.html
talbs: @andy-armstrong, I'll abstract this area out so its less about specific patterns and is more about location of where to put off-canvas UI elements like alerts, modals, and others.

talbs:
> @talbs I know that you say that these templates don't point to real CSS/JS, but I think it would be great if these rendered with the appropriate styling on the doc site. Do you think that subverts their intention slightly, or do you think it would be clear enough to folks what to do. I guess we could have a commented section that says that this brings in the UXPL doc site styling, so switch it out for your own.

@andy-armstrong, as we chatted about in person, these were meant to solve the problem of giving developers a clear set of HTML templates to start new base templates and views in apps from. Specifically, we wanted to give guidance on:

* general HTML structure
* how/where to call assets/references (CSS, JS, critical path vs. supplemental assets)
* needed page/view-based metadata
* general DOM utilities, like skipnav and view/off-view wrappers
* simple starters for more specific view-based needs (primary nav, main content vs. aside content)

Doc-wise, I think:

1. these should be in the UXPL package as files to start from
2. @dsjen's work to create a new landing view for the UXPL Doc Site can reference utilities like this
3. I can add these into our main wiki page (https://github.com/edx/ux-pattern-library/wiki) under "how-tos" or "getting started" sections.

The UXPL Doc Site templates should adhere to the practices in these boilerplates as much as possible. I'll double check that they do now and correct anything as part of this PR.

talbs:
> It might be a good idea to have a right-to-left boilerplate file too.

@andy-armstrong, interesting. I'll try to make a RTL version of our most complex boilerplate (as some things noted there have LTR?RTL implications).

talbs:
@andy-armstrong, I've added a RTL example (https://github.com/edx/ux-pattern-library/blob/talbs/html-boilerplates/pattern-library/html/boilerplate-blank-RTL.html) using our best practices. It turns out that the blank/simple version of the the boilerplate was the right layer to provide this version from. I've included notes and links to more info on RTL within that file too.

talbs:
@andy-armstrong, mind taking another spin through? I've addressed all of your feedback and gave some thoughts to the more admin/doc questions you raised. Thanks for the help!

andy-armstrong:
:+1: Looks great, @talbs.

talbs:
@cptvitamin, I know you're OOO now. Let me know if have time to look this over on Monday (2/1) before I head out. if not, we can figure out some options.

edx-webhook:
Thanks for the pull request, @talbs! I've created [OSPR-1109](https://openedx.atlassian.net/browse/OSPR-1109) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- edx-code email threads
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

nedbat:
@talbs this seems to be near completion, I'll leave it open and delete the OSPR ticket.

andy-armstrong:
@cptvitamin Where did you leave this with Brian? Can we go ahead and merge it as is, or do you plan to do some level of a11y review first? Thanks.

talbs:
@cptvitamin, let me know if you want to chat anything out. I'd love to get this hopefully very close WIP landed for you good folks. Holler back!

andy-armstrong:
@cptvitamin I'd like to merge this story tomorrow as we close out our sprint. Let me know if you think you'll have time to look at it, and if not we can always make a follow-on story to address a11y concerns. Thanks!

pattern-library/html/boilerplate-blank-RTL.html
```diff
@@ -0,0 +1,54 @@
+<!--
+edX Pattern Library: Boilerplate HTML Template: Header + Footer (RTL layout direction!)
+
+NOTES:
+
+* base template with details on how to handle primary/secondary assets, metadata, and basic view-level UI (on canvas vs. conditional and off-canvas elements)
+* right-to-left (RTL) layout configuration have been added
+* see https://github.com/edx/ux-pattern-library/wiki/Styleguide:-HTML#right-to-left-rtl-support for more information
+-->
```
cptvitamin: Should we add an explicit warning that this comment should be removed?  Not sure it still applies, but putting comments before the doctype declaration used to trigger quirks mode.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+        <!-- CASE: if needed, provide a message for older versions of IE -->
+        <!--[if lt IE 9]>
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <div class="nav-skip sr-only">
+            <a class="action-skip" href="#nav-app">Skip to the main navigation</a>
+            <a class="action-skip" href="#content-main">Skip to the main content</a>
+            <a class="action-skip" href="#content-aside">Skip to the aside content</a>
+        </div>
```
cptvitamin: If we use proper sectioning markup/landmark roles, then Skip links are not necessary because screen reader users have another means to move their focus to the main nav, the main content and any aside.  Skip links do help sighted keyboard only users though.  One thing I worry about is maintaining the integrity of these links.  The current platform has many pages where the skip link is there, but the target is not.  I'm in favor here of omitting the skip links entirely and using proper sections/landmarks with logical labels.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-header-app">
+                <header class="header-app" role="banner">
```
cptvitamin: a role of banner is not necessary here.  it is implied by the fact that this is the first header on a page and it is not the child of a section or article.  I would like to see an aria-label on here though: ```aria-label="Site"``` This will be presented to the screen reader user as a "Page banner" landmark (as in the banner for the entire page, that contains page level controls.)  Other acceptable values would be Main, Top or Site.  It often depends on context.  Best to choose something that works in all possible contexts, which is why I think Page is the best choice.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+                        <a class="logo" href="{app or company home url}">
+                            <img class="logo-img" src="{path/to/logo-img.svg}" alt="{app/site name}" />
+                            <!-- example: alternate app/site name (if not a graphical logo) -->
+                            <!-- {app/site name} -->
+                        </a>
+                    </h1>
+                    <!-- insert header content here -->
+
+                    <div class="wrapper-navigation">
+                        <nav class="nav-account" id="nav-account">
```
cptvitamin: this must have an aria-label that describes which navigation section this is (most pages contain multiple nav elements).  An acceptable value here would be ```aria-label="Site"``` since this navigation is for the entire site.  It would be read to the user as "Site navigation".  Other acceptable values would be "Application" or "Main".

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+                                {username}
+                                <span class="sr-only">'s settings/profile</span>
+                            </a>
+                        </nav>
+                    </div>
+                </header>
+            </div>
+
+            <div class="wrapper-navigation">
+                <nav class="nav-app" id="nav-app">
```
cptvitamin: this must have an aria-label that describes which navigation section this is (most pages contain multiple nav elements).  An acceptable value here would be ```aria-label="Course"``` since this navigation is for the user menu.  It would be read to the user as "Course navigation".

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+                    <a href="{url}">Discussions</a>
+                    <a href="{url}">Assignments</a>
+                    <a href="{url}">Assets &amp; Utilities</a>
+                    <a href="{url}">My Progress</a>
+                </nav>
+            </div>
+
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-content">
+                <main class="content-main" id="content-main">
```
cptvitamin: for sectioning/label elements to be meaningful, they need aria-labels, like the above.  I like adding ```aria-label="Main"``` here.  It is presented to the user as "Main content".  Other acceptable values would be "Course" or "Page", although the more specific you get, the more you have to pay attention to the context its being used in.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+                    <!-- insert main content here -->
+                </main>
+
+                <aside class="content-aside" id="content-aside">
```
cptvitamin: This would need an aria-label again.  HArder to put a place holder in here because we don't know what the content in the aside should be.  I also caution against include asides in templates.  There is a lot of confusion around what an aside is.  Many think it is analogous to sidebar, but that is not true.  asides should make sense all by themselves, without the rest of the page content as context.  Many times, what you find in asides, is actually more appropriate for ```<footer>``` elements (the author/bio, the date it was published etc) or simply another nested ```<section>```

pattern-library/html/boilerplate-blank-RTL.html
andy-armstrong: Great point. I've moved all the comments beneath the DOCTYPE to keep them valid. I considered adding a comment to remove it, but that felt redundant as most of the comments in the file should be removed.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+                    <!-- insert aside/supplemental content here -->
+                </aside>
+            </div>
+
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" role="contentinfo">
```
cptvitamin: Since this is the last footer on the page, and not a child of a section or article, we do not need the role of contentinfo here since it is implied.  I would add an aria-label here though. It should be the same as the one used on the header at the top of the page (unless you chose Top of course).  Site or Page would be great.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+                    <!-- insert footer content here -->
+                </footer>
+            </div>
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view">
```
cptvitamin: Since authors are almost always going to need to target this section to manage focus, it would probably be a good idea to give this div an ID attribute.

pattern-library/html/boilerplate-blank.html
```diff
@@ -0,0 +1,51 @@
...
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+            <!-- page/view content is placed here -->
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view">
```
cptvitamin: Since authors are almost always going to need to target this section to manage focus, it would probably be a good idea to give this div an ID attribute.

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,86 @@
...
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" role="contentinfo">
+                    <!-- insert footer content here -->
+                </footer>
+            </div>
+        </div>
+
+        <!-- if needed, wrapper for all off-canvas DOM elements -->
+        <div class="wrapper-off-view">
```
cptvitamin: Since authors are almost always going to need to target this section to manage focus, it would probably be a good idea to give this div an ID attribute.

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,86 @@
...
+            </div>
+
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-content">
+                <!-- insert content here -->
+            </div>
+
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" role="contentinfo">
```
cptvitamin: Since this is the last footer on the page, and not a child of a section or article, we do not need the role of contentinfo here since it is implied. I would add an aria-label here though. It should be the same as the one used on the header at the top of the page (unless you chose Top of course). Site or Page would be great.

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,86 @@
...
+                            <!-- example: alternate app/site name (if not a graphical logo) -->
+                            <!-- {app/site name} -->
+                        </a>
+                    </h1>
+                    <!-- insert header content here -->
+                </header>
+            </div>
+
+            <div class="wrapper-nav">
+                <nav class="nav-app">
```
cptvitamin: this must have an aria-label that describes which navigation section this is (most pages contain multiple nav elements). An acceptable value here would be aria-label="Course" since this navigation is for the user menu. It would be read to the user as "Course navigation".

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,86 @@
...
+        <!--[if lt IE 9]>
+            <p class="alert alert-browserupgrade">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+        <![endif]-->
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+
+            <!-- NOTE: wrapper elements are only needed when they need to manage the grid or larger view/page layout aspects (e.g. spanning the whole viewport) -->
+            <div class="wrapper-header-app">
+                <header class="header-app" role="banner">
```
cptvitamin: a role of banner is not necessary here. it is implied by the fact that this is the first header on a page and it is not the child of a section or article. I would like to see an aria-label on here though: aria-label="Site" This will be presented to the screen reader user as a "Page banner" landmark (as in the banner for the entire page, that contains page level controls.) Other acceptable values would be Main, Top or Site. It often depends on context. Best to choose something that works in all possible contexts, which is why I think Page is the best choice.

pattern-library/html/boilerplate-blank-RTL.html
```diff
@@ -0,0 +1,54 @@
...
+
+<!DOCTYPE html>
+<html class="no-js" lang="en" dir="rtl">
+    <head dir="rtl">
+        <meta charset="utf-8">
+        <!-- always force latest IE rendering engine or request chrome rrame -->
+        <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
```
cptvitamin: If you're aiming for strict validation here, it's best to define these in the headers for the document, not in the page source itself.  IE=edge,chrome=1 is not a valid value for the content attribute here.

pattern-library/html/boilerplate-blank.html
```diff
@@ -0,0 +1,51 @@
...
+
+* base template with details on how to handle primary/secondary assets, metadata, and basic view-level UI (on canvas vs. conditional and off-canvas elements)
+-->
+
+<!DOCTYPE html>
+<html class="no-js" lang="en">
+    <head>
+        <meta charset="utf-8">
+        <!-- always force latest IE rendering engine or request chrome rrame -->
+        <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
```
cptvitamin: If you're aiming for strict validation here, it's best to define these in the headers for the document, not in the page source itself.  IE=edge,chrome=1 is not a valid value for the content attribute here.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,113 @@
...
+* base template with addition of how to start organizing main vs. supplementary/aside content
+* this assumes that the main and aside content are next to each other in the document's flow. This does not neccessarily have to be the case, but the same markup could be used in more complex cases.
+-->
+
+<!DOCTYPE html>
+<html class="no-js" lang="en">
+    <head>
+        <meta charset="utf-8">
+        <!-- always force latest IE rendering engine or request chrome rrame -->
+        <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
```
cptvitamin: If you're aiming for strict validation here, it's best to define these in the headers for the document, not in the page source itself.  IE=edge,chrome=1 is not a valid value for the content attribute here.

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,86 @@
...
+* base template with the additon of how to handle an app-centric header and footer semantically.
+* this assumes that the header and footer need to have their own visual band/are full width against the viewport. If that's not needed, the ``wrapper-*`` elements around them can be removed in favor of the ``wrapper-view`` grid/layout/visual control.
+-->
+
+<!DOCTYPE html>
+<html class="no-js" lang="en">
+    <head>
+        <meta charset="utf-8">
+        <!-- always force latest IE rendering engine or request chrome rrame -->
+        <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
```
cptvitamin: If you're aiming for strict validation here, it's best to define these in the headers for the document, not in the page source itself.  IE=edge,chrome=1 is not a valid value for the content attribute here.

pattern-library/html/boilerplate-blank-RTL.html
andy-armstrong: Given that the Pattern Library has no control over the request, I think it makes sense to include this. I'll add a comment mentioning that this might not work for all browsers, but it should work for most. For example, this well regarded HTML5 boilerplate library includes this:

https://html5boilerplate.com/

In particular:

https://github.com/h5bp/html5-boilerplate/blob/5.3.0/dist/doc/html.md#indexhtml

Does that sound like a reasonable approach?

andy-armstrong: I meant to say I'll also add a comment about adding the header too if possible.

andy-armstrong: Hmmm, this Stack Overflow article says that it is unnecessary for Microsoft browsers:

http://stackoverflow.com/questions/26346917/why-use-x-ua-compatible-ie-edge-anymore

andy-armstrong: It seems that Google Frame is discontinued so the chrome=1 part is unnecessary too!

http://stackoverflow.com/questions/22059060/is-it-still-valid-to-use-ie-edge-chrome-1

I'm going to remove the whole line as you originally recommended. :-)

andy-armstrong:
@cptvitamin Thanks for the detailed feedback. I've addressed your comments and I've also fixed some minor typos and long lines. Please re-review when you have a moment (I've kept my commit separate if you just want to see those changes).

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,122 @@
...
+                    <!-- insert aside/supplemental content here -->
+                </aside>
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" role="contentinfo">
```
cptvitamin: remove the role here and add an aria-label="Page"

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,96 @@
...
+
+        <!-- wrapper for all UI rendered within the page/view -->
+        <div class="wrapper-view">
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-header-app">
+                <header class="header-app" role="banner">
```
cptvitamin: remove the role here and add an aria-label="Page" here

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,96 @@
...
+            <div class="wrapper-content">
+                <!-- insert content here -->
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-footer-app">
+                <footer class="footer-app" role="contentinfo">
```
cptvitamin: Remove role and add an aria-label="Page" here

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,122 @@
...
+                    <a href="{url}">My Progress</a>
+                </nav>
+            </div>
+
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-content">
+                <main class="content-main" id="content-main" aria-label="Main">
```
cptvitamin: @andy-armstrong my bad on this one. You actually want "Content" to be the aria-label here, not Main. I spaced on that one.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,122 @@
...
+                    <!-- insert main content here -->
+                </main>
+
+                <aside class="content-aside" id="content-aside">
```
cptvitamin: Do we want to add an HTML comment here that reminds users to add an aria-label that describes the content/purpose of the aside?

pattern-library/html/boilerplate-header+footer.html
```diff
@@ -0,0 +1,96 @@
...
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-header-app">
+                <header class="header-app" role="banner">
+                    <!-- example: app logo/title -->
+                    <h1 class="header-app-title">
+                        <a class="logo" href="{app or company home url}">
+                            <img class="logo-img" src="{path/to/logo-img.svg}" alt="{app/site name}" />
```
cptvitamin: Since this is an interactive control, the alt text should describe the destination.  Can we append the word "Home" or "Home Page" to {app/site name}

andy-armstrong:
@cptvitamin Please re-review my latest commit when you have a moment.

pattern-library/html/boilerplate-header+footer+content+nav.html
```diff
@@ -0,0 +1,122 @@
...
+            <!--
+              NOTE: wrapper elements are only needed when they need to manage the grid or
+              larger view/page layout aspects (e.g. spanning the whole viewport)
+            -->
+            <div class="wrapper-header-app">
+                <header class="header-app" aria-label="Page">
+                    <!-- example: app logo/title -->
+                    <h1 class="header-app-title">
+                        <a class="logo" href="{app or company home url}">
+                            <img class="logo-img" src="{path/to/logo-img.svg}" alt="{app/site name}" />
```
cptvitamin: @andy-armstrong can you apply the same change you made here:
https://github.com/edx/ux-pattern-library/pull/262/files#diff-eaf72cf3399993215c9656012013d85cR52

Then its a thumbs from me

PR #263, opened 2016-01-22
FIX - Revise LTR/RTL configuration to not use import + if statements

talbs:
We ran into the issue of edx-platform Sass not being able to compile UXPL-based project Sass compilation files due to the following error:

```
Import directives may not be used within control directives or mixins.
```

After looking into things more via the following threads, its best to move away from leveraging importing RTL/LTR support using a Sass ``@if`` clause.

* https://github.com/sass/sass/issues/451
* http://stackoverflow.com/questions/13879042/import-in-if-statement-in-sass
* https://github.com/sass/libsass/issues/1724

- - -

This work revises the LTR/RTL config and settings to not rely on Sass ``@import`` + ``@if`` since Sass doesn't support ``@import`` statements inside of ``@if`` syntax. More details on the steps include: 

* rolled back to using specific bi-app-sass imports for LTR/RTL cases
* updated all UXPL package and Doc Site Sass compilation files to use the new method
* moved if/else -direction variable use to be within ``_grid.scss``

### Reviewers
- [x] Code review: @andy-armstrong 
- [x] Code review: @dsjen 

### Post-review
- [x] Squash commits


pattern-library/sass/patterns/_grid.scss
```diff
@@ -18,6 +18,18 @@
 // ----------------------------
 // #SETTINGS
 // ----------------------------
+// grid - configuration (susy-based)
+
+// set RTL/LTR specific variables
+@if $layout-direction == rtl {
+    $grid-direction-default: rtl;
+    $grid-direction-reversed: ltr;
+
```
dsjen: Nit: extra line

dsjen:
:star2: LGTM :star2:

demo/static/sass/demo-ltr.scss
```diff
@@ -9,6 +9,11 @@
 // ------------------------------
 $layout-direction: ltr;
 
+// ------------------------------
+// #LIB
+// ------------------------------
+@import '../../../node_modules/bi-app-sass/bi-app/bi-app-ltr';
```
andy-armstrong: Nit: I'm wondering if we should add partials for _ltr.scss and _rtl.scss. I'm not super comfortable with every client of the pattern library having to know that we're using bi-app, and it would make it harder to change that dependency. I'm also wondering if we might eventually have some additional actions that we'd perform in these partials, so having them broken out now would make that easier.

Does that make sense? Am I missing something as to why this is a bad idea?

andy-armstrong: In fact, that partial could set the $layout-direction too so that wouldn't have to be repeated.

andy-armstrong:
:+1: Looks great to me.

demo/static/sass/demo-ltr.scss
talbs: Thanks, @andy-armstrong! That's damn good feedback and simplifying this more makes sense.

I've gone ahead and done that with this ``fixup!`` commit (https://github.com/edx/ux-pattern-library/commit/a4dc34ddfc4d5b706391d9557af35d3d5feecd76). The README file compilation examples have also bee updated. How does this sound?

talbs:
@andy-armstrong and @dsjen, I addressed your good feedback. Let me know if you want to take another spin through or if I should move ahead with merging this and using it on our next steps in the platform work.

README.md
```diff
@@ -271,11 +276,12 @@ $layout-direction: ltr;
 @import 'layouts';
 @import 'views';
 @import 'overrides';
-
 ```
 
 **NOTE**: Since both libSass and RubySass lack a way to pass in variables/configuration into their ``@import {file}`` method, each app is responsible for 1) storing any npm-based dependencies, including the edx-pattern-library, in the best directory structure for that app's set up and 2) creating a ``_lib.scss`` partial to import all third party library dependencies from that structure for the Pattern Library to use (see above example).
 
+**NOTE**: We support right-to-left and left-to-right-based layouts. To suppor this within each app, we recommend including a ``_rtl.scss`` or ``_ltr.scss`` file that contains the right layout setting and reference to specific third party utilities.
```
andy-armstrong: Typo: suppor

andy-armstrong: When you say "we recommend", does this mean that you can choose not to include one or the other? Does it matter that ``$layout-direction`` won't be set?

demo/static/sass/demo-ltr.scss
andy-armstrong: Looks great. Thanks!

andy-armstrong:
:+1:

PR #268, opened 2016-01-25
Publish: 01/25/2016

talbs:
* Add missing production CSS from "Removed pointer from progress bar hover." work


talbs:
Bad Talbs!

PR #302, opened 2016-03-01
WIP: added disclosure example

dsjen:
@AlasdairSwan @andy-armstrong -- this is an example of the disclosure component.  I wasn't sure what use cases made sense to include.  Is "Disclosure Component" the right label?

screen shot 2016-03-01 at 6 09 50 pm

### DIFF
--- a/_posts/patterns/2016-03-01-disclosure.md
+++ b/_posts/patterns/2016-03-01-disclosure.md
@@ -1,0 +1,30 @@
+---
+layout:         pattern
+title:          Disclosure Component
+date:           2016-03-01 00:00:00
+
+categories:     patterns
+tags:
+- atomic
+- pattern
+- disclosure
+
+slug:           disclosure
+
+description:    Examples of disclosure component.
+
+info:           TBD
+---
+
+<h3 class="hd-6 example-set-hd">Disclosure Component</h3>
+<div class="example-set">
+    <div class='disclosure' data-expanded-text="Hide Description" data-collapsed-text="Show Description">
+        <button type="button" class="btn-brand disclosure-toggle">
+            Hide Description
+        </button>
+
+        <div class="disclosure-target copy-lead">
+            <p>Raised inside a geyser field, the port of Darkwell is home to vikings lead by Mrs. Chilson. This port wasn't built by a geyser field by accident, as it has rare resources, which is of great importance to the people of Darkwell and its success.</p>
+        </div>
+    </div>
+</div>
--- a/pldoc/static/js/disclosure-example.js
+++ b/pldoc/static/js/disclosure-example.js
@@ -1,0 +1,11 @@
+require(
+    [
+        'edx-ui-toolkit/js/disclosure/disclosure-view.js'
+    ],
+    function(DisclosureView) {
+        'use strict';
+        var view = new DisclosureView({
+            el: '.disclosure'
+        });
+    }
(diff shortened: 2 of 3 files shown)


andy-armstrong:
@dsjen This is awesome! I checked it out locally and it is looking great. Thanks for getting it so close. 

Here are my thoughts:

1. I had the same challenge with the name when creating the wiki status page. I eventually settled on calling it just "Disclosure" as in context it is the disclosure pattern. When we switch to different sections, it will be listed under components, so again I think the "Component" is redundant. Having said this, it is hard to use it as a noun in a sentence, e.g. "just add a disclosure at the top" doesn't work like saying "just add a button at the top".

2. The pattern seems a little bare as it stands. I wonder if the section that is shown and hidden should have some kind of border or spacing to make its extent clearer. @marcotuts do you have any recommendations?

3. It would be great to have an example of the disclosure widget that has an arrow icon like the LMS and Studio ones. I wonder how we can make it clear when to use one over the other. @chris-mike and @marco, what do you guys think?

4. We'll need an introductory paragraph where you have TBD right now. @chris-mike is that something you can help us write again?

5. We should talk with @clrux and @cptvitamin about the a11y of this component. I need to wrap up the work on previewing Jekyll using S3 so that we can all have a URL to play with.

For context, here's the discussion we had about the pattern: https://openedx.atlassian.net/wiki/display/FEDX/Disclosure+Pattern

_posts/patterns/2016-03-01-disclosure.md
```diff
@@ -0,0 +1,30 @@
...
+
+description:    Examples of disclosure component.
+
+info:           TBD
+---
+
+<h3 class="hd-6 example-set-hd">Disclosure Component</h3>
+<div class="example-set">
+    <div class='disclosure' data-expanded-text="Hide Description" data-collapsed-text="Show Description">
+        <button type="button" class="btn-brand disclosure-toggle">
```
cptvitamin: @dsjen we need to add some aria-* here.  First, we need to programmatically indicated the state of the disclosure with ```aria-expanded=true``` when the target is visible and ```false``` when its not.  Second we need to programmatically indicate what is being expanded or collapsed with ```aria-controls``` which takes an IDref.  For that, you will need to add an ID attribute to the sibling ```div``` (in this case).

I noticed that the rendered control already has ```aria-expanded``` but it is lacking the ```aria-controls```.  The markup example in the PL lacks both.

andy-armstrong:
Another thought about naming: other possibilities are disclosure control or disclosure widget. I think I like the former better, but we'd have to get some agreement. There is some precedent with tab controls, although that may be a Windows-centric name.

andy-armstrong:
@clrux FYI this is the disclosure widget story.

clrux:
@dsjen Going back to what Mark was suggesting, let's add a unique `id` to the container that collapses and expands. Then we need to give the toggle button `aria-controls="[unique id]"`. It would be nice too, if the container was programmatically focusable, with focus being sent as a callback to the expanding function *.

* Depending on the use case this may not be that important. But if the expand/collapse toggle button is placed somewhere not immediately before, it would be helpful to send focus so reading can continue intuitively.

marcotuts:
It isn't clear to me where this would be used in the exact visual formulation / design shown in the screenshot above. Are there examples of where this would be used today?

PR #304, opened 2016-03-08
Automate publishing of documentation preview to S3

andy-armstrong:
https://openedx.atlassian.net/browse/FEDX-74


edx-webhook:
## PR Cover Letter
- [ ] A detailed description of the changes in the body of the PR
- [ ] If available, a link to the relevant JIRA ticket in the body of the PR
  - `[FEDX-XXXX](https://openedx.atlassian.net/browse/FEDX-XXXX)`

## Documentation Preview
- [ ] Build a preview site using ``gulp build-preview`` and add the generated link here

## Testing
- [ ] i18n
- [ ] Responsive design
- [ ] RTL

## Reviewers
- [ ] Code review: @Reviewer1
- [ ] Code review: @Reviewer2
- [ ] Documentation review: @...
- [ ] UX review: @...
- [ ] Accessibility review: @...
- [ ] Product review: @...

FYI: Tag anyone who might be interested in this PR here.

## Post-review
- [ ] Squash commits

andy-armstrong:
@fredsmith Can you remind me what the lifecycle is for the S3 bucket. Do the files automatically get removed at some point, or do folks need to perform a manual clean up?

gulp/tasks/jekyll-build-preview.js
```diff
@@ -0,0 +1,29 @@
...
+        gitUtils = require('../util/gitUtils'),
+        childProcess = require('child_process'),
+        previewConfigFile = '_tmp_preview_config.yml',
+        previewSiteDir = '_preview_site';
+
+    gulp.task('jekyll-build-preview', function() {
+        var branch = gitUtils.currentBranch(),
+            previewBaseUrl = '/' + branch + '/';
+        // Create a temporary Jekyll configuration file which specifies the base URL for the preview site
+        childProcess.execSync(
```
dan-f: It looks like [this function can throw an error](https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options). Could you add some error handling?

gulp/tasks/upload-preview.js
```diff
@@ -0,0 +1,17 @@
...
+        previewSiteDir = '_preview_site';
+
+    gulp.task('upload-preview', function() {
+        var branch = gitUtils.currentBranch();
+        childProcess.execSync(
+            'aws s3 sync ' + previewSiteDir + ' s3://' + previewDomain + '/' + branch
+        );
+        console.log('Preview site ready at http://' + previewDomain + '/' + branch);
+    });
+}());
```
AlasdairSwan: @andy-armstrong do we really need the remove and upload preview code in separate files? The require includes at the top of the file are almost identical. It just seems a little intense to have 1 gulp task per file as opposed to a collection of related tasks. Thoughts?

gulp/tasks/jekyll-build-preview.js
dan-f: Actually, checking the Jekyll documentation, it looks like you can simply pass the base url parameter to jekyll via the `--baseurl` argument, removing the need to spawn a child process to create a temporary file.

gulp/tasks/jekyll-build-preview.js
```diff
@@ -0,0 +1,29 @@
...
+
+        // Generate the preview version of the site
+        console.log('Generating preview for branch ' + branch);
+        childProcess.execSync(
+            'jekyll build --config _config.yml,' + previewConfigFile + ' --destination ' + previewSiteDir
+        );
+
+        // Invoke WebPack to rebuild the JavaScript in the preview site with the correct base URL
+        childProcess.execSync(
+            'OUTPUT_ROOT=' + previewSiteDir + '/ SITE_ROOT=/' + branch + '/ ./node_modules/.bin/webpack'
```
dan-f: You can reuse `previewBaseUrl` here.

gulp/tasks/upload-preview.js
```diff
@@ -0,0 +1,17 @@
+(function() {
+    'use strict';
+
+    var gulp = require('gulp'),
+        gitUtils = require('../util/gitUtils'),
+        childProcess = require('child_process'),
+        previewDomain = 'ux-test.edx.org',
```
dan-f: What if we were to parameterize `previewDomain`? This would open up the previewing workflow to non-internal contributors.

gulp/tasks/upload-preview.js
```diff
@@ -0,0 +1,17 @@
...
+        previewSiteDir = '_preview_site';
+
+    gulp.task('upload-preview', function() {
+        var branch = gitUtils.currentBranch();
+        childProcess.execSync(
+            'aws s3 sync ' + previewSiteDir + ' s3://' + previewDomain + '/' + branch
```
dan-f: Similar point to above - perhaps we could refactor this command into a config file or command line argument such that folks using non-s3 for hosting can upload previews as well.

AlasdairSwan:
@andy-armstrong I have some general comments around the S3 preview created

1. Is it password protected?
2. How long will the preview exist for?
3. Are the preview pages crawlable?

andy-armstrong:
@AlasdairSwan I spoke to @fredsmith about this. The S3 bucket is not password protected, has no expiration date, but is not crawlable without knowing the particular URL used. His recommendation is to add an expiration (30 days seems reasonable) but to otherwise leave it as is. How does that sound?

gulp/tasks/jekyll-build-preview.js
andy-armstrong: @dan-f Unfortunately the ``--baseurl`` parameter is to the Jekyll serve command but not to build (that's what I tried first). That's why I had to go with this somewhat ugly hack of a temporary file.

It's an interesting point about adding error handling to this function. My thinking is that if there is an error then the overall gulp task will fail with an error message. I think adding error handling to every command will make the code much harder to understand.

I'm starting to think that Gulp is a bit heavyweight and that this would have all been easier with a makefile. You don't typically have to write error handling in makefiles. You just chain together Unix commands. You also don't have to mess around with Gulp plugins for every task you want to do.

gulp/tasks/upload-preview.js
andy-armstrong: @AlasdairSwan I thought that there was a convention of one file per task, but I'm happy to relax that if folks think it is clearer. @dan-f, what do you think?

gulp/tasks/upload-preview.js
andy-armstrong: That's an interesting question. We could make it an environment variable that everyone has to set (since you already have to set AWS environment variables).

gulp/tasks/upload-preview.js
andy-armstrong: I'm having a harder time understanding what this would look like. How would they plug in their own hosting solution? I'd rather wait until we have a use case to know what this might look like. For now, I think these tasks are just for edX to use internally, which I suppose is a little unclean in an open source repo. If I removed the edx.org constant then having a dependency on S3 seems okay for now. I could rename it upload-s3-preview if you think that's clearer.

gulp/tasks/jekyll-build-preview.js
dan-f: Well FWIW, the commands could fail whether initiated from the a Makefile, shell script, python script, or gulp task.

I actually feel like surrounding this with a try/catch and console.logging something might help readability.

gulp/tasks/upload-preview.js
dan-f: I'm not super familiar with gulp patterns, so I don't have a strong preference. Can we settle on something relatively standard?

andy-armstrong: I think the convention was just something that Brian came up with. I've pushed up a new commit that merges everything into one file and IMO it is much easier to reason about.

andy-armstrong:
@AlasdairSwan @dan-f I've addressed your feedback. Please re-review when you have a moment. Thanks.

gulp/tasks/preview.js
```diff
@@ -0,0 +1,72 @@
...
+        // Generate the preview version of the site
+        console.log('Generating preview for branch ' + branch);
+        childProcess.execSync(
+            'jekyll build --config _config.yml,' + previewConfigFile + ' --destination ' + previewSiteDir
+        );
+
+        // Invoke WebPack to rebuild the JavaScript in the preview site with the correct base URL
+        childProcess.execSync(
+            'OUTPUT_ROOT=' + previewSiteDir + '/ SITE_ROOT=' + previewBaseUrl + ' ./node_modules/.bin/webpack'
+        );
```
dan-f: Can we clean up (delete) `previewConfigFile` since it's not needed at this point?

gulp/tasks/preview.js
```diff
@@ -0,0 +1,72 @@
...
+            'jekyll-build-preview',
+            'upload-preview',
+            callback);
+    });
+
+    gulp.task('jekyll-build-preview', function() {
+        var branch = gitUtils.currentBranch(),
+            previewBaseUrl = '/' + branch + '/';
+        // Create a temporary Jekyll configuration file which specifies the base URL for the preview site
+        childProcess.execSync(
```
dan-f: Still wondering about `try`/`catch`.

andy-armstrong: Sorry that that wasn't clear. Each gulp task already catches errors and displays them, so having our own try/catch seems IMO verbose and unnecessary. I've looked at a lot of gulp task implementations using Google and I don't see any that do explicit error handling. As I said, gulp is more like a makefile, and in makefiles you don't write error handling.

I tried forcing some errors and while the exceptions could be clearer in some cases, the actual error is always reported. For example, here's what happens when I switch the command from 'aws' to 'awsx':

```
$ gulp upload-preview
[23:44:43] Using gulpfile ~/edx/src/ux-pattern-library/gulpfile.js
[23:44:43] Starting 'upload-preview'...
/bin/sh: awsx: command not found
[23:44:43] 'upload-preview' errored after 14 ms
[23:44:43] Error: Command failed: awsx s3 sync _preview_site s3://ux-test.edx.org/andya/s3-preview
/bin/sh: awsx: command not found

    at checkExecSyncError (child_process.js:1346:13)
    at Object.execSync (child_process.js:1386:13)
    at Gulp.<anonymous> (/Users/andya/edx/src/ux-pattern-library/gulp/tasks/preview.js:54:26)
    at module.exports (/Users/andya/edx/src/ux-pattern-library/node_modules/gulp/node_modules/orchestrator/lib/runTask.js:34:7)
    at Gulp.Orchestrator._runTask (/Users/andya/edx/src/ux-pattern-library/node_modules/gulp/node_modules/orchestrator/index.js:273:3)
    at Gulp.Orchestrator._runStep (/Users/andya/edx/src/ux-pattern-library/node_modules/gulp/node_modules/orchestrator/index.js:214:10)
    at Gulp.Orchestrator.start (/Users/andya/edx/src/ux-pattern-library/node_modules/gulp/node_modules/orchestrator/index.js:134:8)
    at /usr/local/lib/node_modules/gulp-cli/lib/versioned/^3.7.0/index.js:46:20
    at process._tickCallback (node.js:355:11)
    at Function.Module.runMain (module.js:503:11)
```

gulp/tasks/preview.js
andy-armstrong: Good call. I've pushed up a change for this.

AlasdairSwan:
30 days sounds reasonable, looks good to me :+1:

gulp/tasks/preview.js
dan-f: I see. Thanks for clarifying.

dan-f:
:+1:

PR #303, opened 2016-03-02
Alasdair/adding dropdown menu example

AlasdairSwan:
https://openedx.atlassian.net/browse/FEDX-36

PR #305, opened 2016-03-11
FEDX-103 Added styling to dropdown menu, a JS tab to the pldoc site a…

AlasdairSwan:
…nd documentation for the dropdowm menu element
@andy-armstrong I pulled this out into a new branch to resolve the git gremlins
Previous PR (with review comments) https://github.com/edx/ux-pattern-library/pull/303

Sandbox url: http://ux-test.edx.org/alasdair/FEDX-103-dropdown-menu/


AlasdairSwan:
@andy-armstrong can you please review this update which addresses the comments you made in https://github.com/edx/ux-pattern-library/pull/303

AlasdairSwan:
@clytwynec can you please review this dropdown component from an a11y point of view? I took a look at your recent PR to edx-platform and tried to incorporate the changes.
@cahrens can you please review from a UX point of view?

cahrens:
@AlasdairSwan did you mean to tag @chris-mike?

AlasdairSwan:
Sorry @cahrens, I did mean to tag @chris-mike.

clytwynec:
@AlasdairSwan  It looks like focus is trapped in the menu (until Esc or Space) if using arrow keys to navigate, but if tabbing the user can tab past the end or out the top of the menu without closing it.  Would it be possible to add the behavior that you have for arrows in the dropdown to tabs?

AlasdairSwan:
Thanks for catching that @clytwynec, I will update to trap tabs too.

chris-mike:
@AlasdairSwan Since we're using a "link" affordance in the dropdown, I wrestled with whether the color was too light and the on/off hover contrast too weak. Comparing to what is live currently, it looks like we have three different designs (subtle differences between drupal, platform, lms). 

1. Can you reverse the colors, so the default is black (same as the "name" that appears left of the arrow) and hover is blue 
2. Can you add an additional "underline" action on hover?

In addition, in your "already rendered DOM" example, can you make the menu wider so that nothing is at 2 lines. I know it's just for show, but I don't like the idea of multi-line listing.

Thank you.

chris-mike:
PS: I realize that by adding an underline on hover, we may need to adjust the pattern library's "link" CSS as well. @andy-armstrong

clrux:
FWIW I also think we need underline on hover. I'm a huge fan of two methods of change: color, and in case color isn't discernible, an underline of sorts. We have a lot of dark blue links that turn darker blue (black?) when hovered. It's really hard to determine. If this is work that needs to get into the UXPL I'm happy to work on that; realizing this may not be the appropriate place or scope.

AlasdairSwan:
@clytwynec I have updated the sandbox with the tab trap fix http://ux-test.edx.org/alasdair/FEDX-103-dropdown-menu

AlasdairSwan:
@chris-mike @clrux I will update to use underline on hover, but will limit the scope of that change to just this element. There is currently no specific link section in the UXPL docs, I think there should probably be a specific story for addressing links across the whole pattern library to ensure scope creep doesn't prevent this current PR from merging. @andy-armstrong thoughts?

_posts/patterns/2016-03-01-dropdown-menu.md
```diff
@@ -0,0 +1,54 @@
...
+<div class="example-set">
+	<div class="js-user-cta"></div>
+</div>
+
+<h4 class="hd-6 example-set-hd">Dropdown with Image</h4>
+<div class="example-set">
+	<div class="js-user-cta-img"></div>
+</div>
+
+<h4 class="hd-6 example-set-hd">Already Rendered DOM</h4>
```
andy-armstrong: Nit: maybe ``Pre-Rendered Dropdown``.

andy-armstrong: I love having this example, btw. It makes it very clear how to do both approaches to rendering the dropdown.

_posts/patterns/2016-03-01-dropdown-menu.md
```diff
@@ -0,0 +1,54 @@
...
+<div class="example-set js-dropdown-menu-vol-4">
+	<div class="wrapper-more-actions dropdown-menu-container">
+		<a href="/" class="menu-title">Vol 4</a>
+		<button type="button" class="button button-more has-dropdown js-dropdown-button" aria-haspopup="true" aria-expanded="false" aria-controls="edx-user-menu">
+		    <span class="sr-only">User options dropdown</span>
+		</button>
+		<ul class="dropdown-menu is-hidden" id="edx-user-menu" tabindex="-1">
+	        <li class="dropdown-item"><a href="/" class="action">Wheels Of Confusion</a></li>
```
andy-armstrong: Nit: the indentation is slightly off here.

_posts/patterns/2016-03-01-dropdown-menu.md
```diff
@@ -0,0 +1,54 @@
...
+</div>
+
+<h4 class="hd-6 example-set-hd">Dropdown with Image</h4>
+<div class="example-set">
+	<div class="js-user-cta-img"></div>
+</div>
+
+<h4 class="hd-6 example-set-hd">Already Rendered DOM</h4>
+<div class="example-set js-dropdown-menu-vol-4">
+	<div class="wrapper-more-actions dropdown-menu-container">
```
andy-armstrong: Nit: use four spaces instead of tabs.

package.json
```diff
@@ -29,9 +29,10 @@
     "susy": "^2.2.9"
   },
   "devDependencies": {
+    "backbone": "*",
     "browser-sync": "*",
     "del": "*",
-    "edx-ui-toolkit": "~0.3.0",
+    "edx-ui-toolkit": "https://github.com/edx/edx-ui-toolkit.git#4bc0a3ac2ef92f5ee044aaab9230826542798979",
```
andy-armstrong: Just a reminder not to merge until we've done a UI Toolkit release and this can refer to a version number.

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+// #UTILITIES
+// ----------------------------
+
+
+// ----------------------------
+// #GENERAL
+// ----------------------------
+
+.dropdown-menu-container {
+  height: 40px;
```
andy-armstrong: It seems a little off that this is hard-coded in pixels. However, when I resize the window everything seems to work fine, so maybe that's not an issue. Is ``40px`` right for all browsers including mobile?

chris-mike:
@AlasdairSwan pointed out the block-level list styling in the UXPL (see lists > "list with visual dividers"). We're going to implement block-level styling (linkable, no background color) with the text color change (black default, blue on hover), and add an underline on hover.

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+
+.menu-title {
+  font: {
+    family: $font-family-sans-serif;
+    size: $dropdown-menu-button-font-size;
+    weight: font-weight(bold);
+  }
+  color: $dropdown-menu-font-color;
+  margin: {
+    left: 15px;
```
andy-armstrong: I think we should look at replacing all of the hard-coded pixel counts in here with parameterized values. You should talk with @dsjen who had talbs's help when he created alerts. See for example:

https://github.com/edx/ux-pattern-library/blob/master/pattern-library/sass/patterns/_alerts.scss#L99

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: Interesting point @andy-armstrong. In the marketing site the dropdown is hidden for smaller devices and in edx-platform there aren't any responsive pages that use it so a mobile-friendly version of the component isn't in place. I'm not sure we want to reduce the font-size on mobile because then it would be difficult to press the links. My recommendation is that for responsive the element flows to its own line (so the Sass for that would be unique to the implementation and would not be a part of the UXPL). @chris-mike does that sound reasonable?

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+
+// ----------------------------
+// #CONFIG
+// ----------------------------
+$dropdown-menu-button-font-size: font-size(small) !default;
+$dropdown-menu-list-font-size: font-size(base) !default;
+$dropdown-menu-z-index: 1001 !default;
+$dropdown-menu-button-text-align: left !default;
+$dropdown-menu-list-text-align: left !default;
+$dropdown-menu-width: 170px !default;
```
andy-armstrong: Nit: another hard-coded pixel width

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+  z-index: $dropdown-menu-z-index;
+  border-radius: 4px;
+  box-shadow: 0 2px 24px 0 rgba(0,0,0,0.3);
+  border: 1px solid $dropdown-menu-border-color;
+  margin-top: 0;
+  padding: 5px 10px;
+  width: $dropdown-menu-width;
+
+  &:before {
+    @include transform( rotate(-45deg) );
```
andy-armstrong: How does the dropdown button's triangle work? Should it be using the edX icon font for accessibility?l

andy-armstrong:
@AlasdairSwan this is looking awesome! I think we need to work out how best to size this using pre-existing size variables rather than hard-coded integers. It is also important that this work on mobile so you should tag someone from that team (probably Akiva) when you have a solution.

clytwynec:
@AlasdairSwan That change works great. Thanks! :+1:

pattern-library/sass/patterns/_dropdown-menu.scss
chris-mike: Sure, that's reasonable @AlasdairSwan .

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: Interestingly the pattern library doesn't have a caret icon, which is why I used this. The possibility should exist for one of the icons from the library to be used 'tho. Perhaps it could default to the CSS rendered caret (which is a11y) and the view could accept an optional pattern library icon name which would replace the default with the icons. Thoughts @andy-armstrong? Also, I just noticed that this isn't the code for the button arrow, it's for the white background of the dropdown menu, but the question still stands re. how to integrate UXPL icons with this element.

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+  background: none;
+  width: 30px;
+  height: 30px;
+  padding: 0;
+  margin-top: 6px;
+  border: none;
+  vertical-align: top;
+  position: relative;
+
+  &:after {
```
AlasdairSwan: @andy-armstrong this is the button's down caret

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+// #INDIVIDUAL CASES
+
+
+// ----------------------------
+// #CONFIG
+// ----------------------------
+$dropdown-menu-button-font-size: font-size(small) !default;
+$dropdown-menu-list-font-size: font-size(base) !default;
+$dropdown-menu-z-index: 1001 !default;
+$dropdown-menu-button-text-align: left !default;
```
dsjen: I believe to support RTL, `left` should be enclosed in with text-align, e.g. `text-align(left)`.

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+$dropdown-menu-list-text-align: left !default;
```
dsjen: Same `text-align(left)` suggestions.

pattern-library/sass/patterns/_dropdown-menu.scss
dsjen: Hmm, maybe `spacing-vertical(large)`?

dsjen: Actually, I suppose the RTL version of this would be `@include left(spacing-vertical(large))`

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+.menu-title {
+  font: {
+    family: $font-family-sans-serif;
+    size: $dropdown-menu-button-font-size;
+    weight: font-weight(bold);
+  }
+  color: $dropdown-menu-font-color;
+  margin: {
+    left: 15px;
+    right: 2px;
```
dsjen: The RTL version would be `@include right(2px)`.  :)

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+  }
+  line-height: 40px;
```
dsjen: It seems like 40px is used a few times.  Would it make sense to pull into a variable?

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+.menu-image {
+  @include margin-right(6px);
+  width: 40px;
+  height: 40px;
+}
+
+.button {
```
dsjen: Would this affect all `button` classes?  Maybe `menu-button` instead?

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+  background: $dropdown-menu-background-color;
+  z-index: $dropdown-menu-z-index;
+  border-radius: 4px;
+  box-shadow: 0 2px 24px 0 rgba(0,0,0,0.3);
+  border: 1px solid $dropdown-menu-border-color;
+  margin-top: 0;
+  padding: 5px 10px;
+  width: $dropdown-menu-width;
+
+  &:before {
```
dsjen: I didn't realize this was initially for the caret.  A comment would be enlightening!

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+  }
+
+  &.is-hidden {
+    display: none;
+  }
+}
+
+// ----------------------------
```
dsjen: Is this left over from copy/paste?

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,173 @@
...
+  height: 40px;
+  position: relative;
+  display: inline;
+}
+
+.dropdown-menu {
+    list-style-type: none;
+}
+
+.menu-title {
```
dsjen: This is sort of a philosophical question.  Should just about every class be prefaced by `dropdown-menu`?  It would namespace the components, although `dropdown-menu` is rather lengthy...

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: I used text-align at the element level to make overriding defaults easier (i.e. you don't need to remember to include the text-align wrapper there). Is the UXPL pattern to using text-align at the variable level?

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: Good point, I think I should probably wrap all of these classes in a ```.dropdown-menu-wrapper``` class to keep the scope to the component.

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: Agreed, I must have been sleep-coding when I did this one :)

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: Yep

pattern-library/sass/patterns/_dropdown-menu.scss
dsjen: Oh, I see!  I'm not sure what the preferred approach would be.  I guess I did find this confusing, but it's up to you.

pattern-library/sass/patterns/_dropdown-menu.scss
dsjen: I cannot imagine why you'd be sleep deprived!

pattern-library/sass/patterns/_dropdown-menu.scss
dsjen: Makes sense!

dsjen:
I'm having a very odd build error `fatal: bad object 4bc0a3ac2ef92f5ee044aaab9230826542798979` and I think because of that when I load the page, I get a `Uncaught Error: Cannot find module "backbone"` JS error in the console.  Are others able to run this without issue?

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: I agree that it's confusing, I will take a look into the UXPL code and see if there is a pattern already established. Thanks for calling this out, it will be good to agree on the approach for more code is written and confusion built into the UXPL.

dsjen: Ok, thanks!

pattern-library/sass/patterns/_dropdown-menu.scss
AlasdairSwan: I think this is actual okay in the context. I am setting it as a variable so it can be easily overridden and with what is being done with the styles I don't think it would be possible to achieve without an explicit width being set. Note that in edx-platform the menu is currently set to ```width: 170px;```.

AlasdairSwan:
@chris-mike I have updated the sandbox (http://ux-test.edx.org/alasdair/FEDX-103-dropdown-menu) can you please take a look?

@andy-armstrong @dsjen I have addressed your comments, can you please review?

pattern-library/sass/patterns/_dropdown-menu.scss
```diff
@@ -0,0 +1,166 @@
...
+// ----------------------------
+$dropdown-menu-title-height: 40px !default;
+$dropdown-menu-button-height: calc(#{$dropdown-menu-title-height} * 0.75) !default;
+$dropdown-menu-button-top-offset: 6px !default;
+$dropdown-menu-image-title-spacer: 6px !default;
+$dropdown-menu-list-spacing-base: spacing-vertical(x-small) !default;
+$dropdown-menu-button-font-size: font-size(small) !default;
+$dropdown-menu-list-font-size: font-size(base) !default;
+$dropdown-menu-z-index: z-index(very-front) !default;
+$dropdown-menu-button-text-align: left !default;
```
dsjen: It doesn't look like this variable is used anywhere.

dsjen:
I left one comment about a sass variable that I believe is unused now.  Also, I noticed when I select an item in the dropdown examples that it makes me to a 404 page.  I believe that would only happen on the test environment though, but just wanted to point that out.

Very exciting PR!  LGTM!  :+1:

AlasdairSwan:
@dsjen yeah my example is just not very good and uses '/' for links.

andy-armstrong:
:+1: Awesome stuff, @AlasdairSwan!

chris-mike:
@AlasdairSwan The DOM example does not use block-level styling. Is this intentional?  Otherwise, everything looks great.

AlasdairSwan:
@chris-mike yes, that was to show that different styles can be applied and the dropdown still works. Thanks for the review.

PR #308, opened 2016-03-22
Publish docs of v0.11.0

AlasdairSwan:
@andy-armstrong please review


andy-armstrong:
:+1: That was a most boring review :-) Maybe we don't need doc PR reviews

PR #311, opened 2016-03-24
Alasdair/publish 03 24 2016

AlasdairSwan:
@andy-armstrong @dsjen please review - this is around https://openedx.atlassian.net/browse/FEDX-124. I think there may have been an issue with the previous PR I created not actually including the updates.


.gitignore
```diff
@@ -33,6 +33,7 @@ nbproject
 .sass-cache
 *.map
 _site
+_tmp_preview_config.yml
```
dsjen: I think I'm out of the loop.  What is this file?

AlasdairSwan: I think this is to generate the S3 sandbox version, is that correct @andy-armstrong ?

dsjen: Ah!  I'm mostly just curious. :)

dsjen:
LGTM!

.gitignore
andy-armstrong: Yes, I had to create a temporary configuration file to get the preview to work. Not ideal, but oh well.

andy-armstrong:
:+1: Fingers crossed that this will fix it! Is there anything you can add to the wiki page to clarify the process or to help with troubleshooting?

AlasdairSwan:
@andy-armstrong assuming it fixes it I can add something around making sure the changes are actually present in the PR and about checking the settings tab to ensure the build was successful (in this case it was successful, it was just not the correct code).

AlasdairSwan:
@andy-armstrong the build worked, and it updated ux.edx.org pretty much instantly.

dsjen:
:balloon: congrats!!!  I see it...so pretty.

PR #312, opened 2016-03-31
Updating Modernizr; adding FontAwesome to icon init

clrux:
#[UXPL-33](https://openedx.atlassian.net/browse/UXPL-33)

This work relates to [AC-329](https://github.com/edx/edx-platform/pull/11438) in that in order to get the icon fonts loading properly for the video player, which currently uses AFontGarde and its methods, both `Modernizr` and `edx-icons` needed to be modified on the UXPL side.

### Modernizr

The instance of Modernizr previously used in the UXPL was throwing JavaScript errors in the platform relating to `offsetHeight()`. I included `gulp-modernizr` and its workflow to generate the Modernizr build when the site loads (for development, not for production). You can modify which supports are included or excluded in the `modernizr-config.json` file.

### edx-icons

The UXPL uses a custom set of icon fonts that we've dubbed "edx-icons", but the platform uses "fontawesome". When we load the UXPL into the platform, we need to be able to support "edx-icons" for future work, but also "fontawesome" for all the existing work.

Once we fully integrate the UXPL, replacing Font Awesome with edx-icons, we can drop support for "fontawesome" in this file.

## Reviewers

- [x] @andy-armstrong 
- [x] @bjacobel


bjacobel:
:thumbsup:

500 lines of vendored code gone!

bjacobel:
Just realized the 500 -lines are because the new Modernizr build is minified/ Since our pipeline does minification on its own, does Modernizr need to be minified?

clrux:
@bjacobel Unsure. When you use the [Modernizr](http://modernizr.com) build tool, it spits out minified.

bjacobel:
Looks like you can un-check here to minify.

modernizr build tool page

clrux:
Pfft, didn't even see that. Thanks. Do we want to un-minify it?

bjacobel:
Yeah, I think so, for two reasons:
- better diffing support when we need to upgrade it it's more obvious how much of a change it is
- ability to use unminified code in development.

clrux:
Added.

andy-armstrong:
@clrux What do you think of the idea of using the Modernizr command line tool to automatically build the latest version as part of the Pattern Library gulp tasks? This automation seems better than requiring you (or someone else) to generate a Modernizr download manually and then check it in.

You can read about their command line here:

https://modernizr.com/docs

pattern-library/js/edx-icons.js
```diff
@@ -1,9 +1,13 @@
...
-        'use strict';
+define(['afontgarde'], function() {
+    'use strict';
 
-        window.AFontGarde('edx-icons', {
-            // Check a few random icons to see if our font loads
-            glyphs: '\uE621\uE622\uE623'
-        });
+    // Old icon fonts, currently used in the platform.
+    window.AFontGarde('FontAwesome', {
```
andy-armstrong: I guess I misunderstood you here @clrux. I thought that the Font Awesome icons addition would only be in the platform. I don't love the idea of having this codified in the Pattern Library too.

clrux:
I'll look into it!

pattern-library/js/edx-icons.js
clrux: It would only be in platform, but this file IS in the platform, yes? Do we need two separate versions then - one for the UXPL and another for the platform?

andy-armstrong: I think of platform as being files in edx-platform. This file will be used by any edX IDAs that want consistent styling, e.g. Insights.

andy-armstrong: So yes, I do think you should have an edx-platform specific version that lives in that repo. It could import this one and then override it, or it could just redefine it from scratch.

clrux:
@andy-armstrong It appears this PR is holding up AC-329 as Modernizr is throwing an error. For the sake of that PR, would you mind if I just updated Modernizr here, and did another PR that added it into the Gulp flow as mentioned above? I'd really love to merge 329 and it's been out for months.

andy-armstrong:
@clrux It is only a couple of lines to do this the right way. I can do it for you if you tell me the Modernizr settings that we use.

andy-armstrong:
@clrux Also, please move the Font Awesome change into the edx-platform repo.

clrux:
@andy-armstrong See options here: https://github.com/edx/ux-pattern-library/pull/312/files#diff-8e027a74948e9b6336b51251cee26dc0R1

clrux:
Also, I was sorta working on this on Friday and planned on picking it back up this week. I believe the current setup is generating the custom built Modernizr file, but it's not yet putting it in the right place. Are you going to take this, or should I continue?

andy-armstrong:
@clrux Why don't you take it as far as you can and then tag me with the remaining problems are. Thanks.

clrux:
@andy-armstrong @bjacobel This adds Modernizr on the Gulp flow, then adds it to the overall compile. Settings are in the JSON file.

Please review asap as this is holding up AC-329. Thanks!

andy-armstrong:
:+1: Looks great @clrux. I'll give you a pass on adding Font Awesome into the Pattern Library as it shouldn't affect folks and we can remove it again if we don't document it.

PR #313, opened 2016-04-01
Clean up library import logic

andy-armstrong:
Clean up the way that library imports are handled, to make it easier on consumers. Now the assumption is that the Pattern Library's node_modules directory will be added to the Sass load path, which means that references can be relative and can work in any client.

@AlasdairSwan @bjacobel @dan-f please review.

### DIFF
--- a/pattern-library/sass/global/_lib.scss
+++ b/pattern-library/sass/global/_lib.scss
@@ -1,0 +1,8 @@
+// ------------------------------
+// edX Pattern Library: Third Party Libraries
+
+// About: third party libraries and dependencies import
+
+@import 'bourbon/app/assets/stylesheets/bourbon';
+@import 'susy/sass/susy';
+@import 'breakpoint-sass/stylesheets/breakpoint';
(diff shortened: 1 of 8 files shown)


pattern-library/sass/global/_lib.scss
```diff
@@ -0,0 +1,8 @@
+// ------------------------------
+// edX Pattern Library: Third Party Libraries
+
+// About: third party libraries and dependencies import
+
+@import 'bourbon/app/assets/stylesheets/bourbon';
+@import 'susy/sass/susy';
+@import 'breakpoint-sass/stylesheets/breakpoint';
```
dan-f: Ahhh :)

dan-f:
Looking great @andy-armstrong! Does this mean that consumers of the UXPL (edx-platform, for example) can now import partials such as `global/_lib.scss`?

dan-f:
I now see that the answer to my above question is "yes" via https://github.com/edx/edx-platform/pull/11957#issuecomment-204243831

AlasdairSwan:
Nice updated @andy-armstrong :+1:

bjacobel:
Much cleaner :thumbsup:

PR #318, opened 2016-04-07
Fixing Modernizr; don't need previous config file

clrux:
## Notes

The previously merged PR successfully added Modernizr into the Gulp workflow, but the configuration JSON file was unnecessary. Apparently it came from somewhere else when I was trying to get this working.

Options for the Modernizr build go into the Gulp task, and should now include the correct build parameters.

### Reviewers

- [x] @andy-armstrong


andy-armstrong:
:+1:

PR #320, opened 2016-04-08
Ensuring icons work with or without RequireJS

clrux:
## Notes

To make the edx-icons available for use in both LMS and Studio, I needed to ensure that the file loads with or without RequireJS. This work makes that happen.

## Reviewer

- [x] @andy-armstrong OR
- [ ] @bjacobel OR
- [ ] @AlasdairSwan

### DIFF
--- a/pattern-library/js/edx-icons.js
+++ b/pattern-library/js/edx-icons.js
@@ -1,13 +1,18 @@
-define(['afontgarde'], function() {
+;(function(define) {
     'use strict';
-
-    // Old icon fonts, currently used in the platform.
-    window.AFontGarde('FontAwesome', {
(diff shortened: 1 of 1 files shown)


andy-armstrong:
:+1: if you've tested it against platform to be sure it does what you need.

FYI a more generic solution would be to model this on how other JavaScript libraries work, such as Backbone:

https://github.com/jashkenas/backbone/blob/master/backbone.js#L8

Your solution will only work inside edx-platform, but I can't imagine others will need it so I think that's okay.

clrux:
Thanks Andy. I tested it within platform and it works on both LMS and Studio! Release PR coming shortly...

PR #322, opened 2016-04-14
Update Pattern Library Ruby libraries

andy-armstrong:
I've updated the Pattern Library Ruby libraries (``bundle update``) as @clrux was having difficulty running today.

@AlasdairSwan @clrux @bjacobel please review.


bjacobel:
:thumbsup:

PR #323, opened 2016-04-20
Updating UXPL icons to not use AFG

clrux:
## 1 of 4

This is part one of four that updates the UX Pattern Library to use the base icon font set, removing AFontGarde from the mix. Tests and conversation deemed AFontGarde unnecessary so this set of work aims to simply the library be removing unneeded dependencies and markup.

### Parts 2, 3, and 4

The other three parts can come at a later date, when other workloads calm down. They are as follows:

* remove the afontgarde use from the video player on edx-platform
* remove afontgarde from the pattern library
  * afontgarde.js
  * edx-icons.js
  * roll back Modernizr to the previous version
* remove afontgarde from edx-platform
  * package.json version bump to pick up the uxpl changes
  * removing edx-icons from the video player dependencies
  * remove edx-icons and afontgarde from require loaders

### Reviewers 

- [ ] @andy-armstrong

### DIFF
--- a/_posts/patterns/2015-04-15-icons.md
+++ b/_posts/patterns/2015-04-15-icons.md
@@ -37,25 +37,15 @@ info:               Icons are used to clearly and visually represent some inform
 
     <div class="example example-icon">
         <div class="icon-display">
-            <span class="icon-fallback icon-fallback-text">
-                <span class="icon icon-error" aria-hidden="true"></span>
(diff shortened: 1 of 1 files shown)


bjacobel:
This is superseded by https://github.com/edx/ux-pattern-library/pull/340, right? Can it be closed?


cc @alisan617

PR #325, opened 2016-05-03
Remove imagemin from the gulp pipeline

andy-armstrong:
## Description

[FEDX-178](https://openedx.atlassian.net/browse/FEDX-178)

This change removes imagemin which is incompatible with the latest versions of our libraries, and which isn't used in the current Pattern Library anyway (as we don't publish images).

I then fixed a bug with compiling the sass with newer versions of libsass:

https://github.com/sass/libsass/issues/1746

## Reviewers
 - [ ] @bjacobel


bjacobel:
:thumbsup:

PR #327, opened 2016-05-06
Add a Travis build

andy-armstrong:
## Description

Add a Travis configuration to do a production build. There are no tests at this point, but the intention is just to make sure that the library itself can still be packaged.
  
## Post-review

## Reviewers
- [x] @bjacobel


andy-armstrong:
@bjacobel It seems that the Pattern Library sass build is running out of memory on Travis. I'm not sure how best to fix this other than to try switching to Webpack... any ideas?

andy-armstrong:
@bjacobel @dsjen @dan-f @AlasdairSwan please review this one too. I've switched over to using Webpack instead of gulp-sass primarily so that Travis can build without running out of memory (although we wanted to move to Webpack anyway). 

In order to land this, I also had to remove the precompiled CSS from the Pattern Library package, as Webpack cannot generate CSS that includes relative paths. From talking with @talbs, it doesn't seem that anyone should be using the precompiled CSS, as they should instead be importing the partials.

bjacobel:
:thumbsup:

bjacobel:
Just a question - why (in the `travis.yml`) does this build need Chrome and Firefox running?

andy-armstrong:
@bjacobel Great question. That was bad cut-and-paste from the UI Toolkit Travis configuration. I've removed them and also tried @dan-f's suggested way to specify multiple languages.

gulp/tasks/pldoc.js
```diff
@@ -1,20 +1,13 @@
...
-    handleErrors = require('../util/handleErrors'),
-    sass = require('gulp-sass'),
-    sourcemaps = require('gulp-sourcemaps'),
-    rename = require('gulp-rename'),
-    size = require('gulp-filesize'),
-    webpack = require('gulp-webpack'),
-    ghPages = require('gulp-gh-pages');
+    webpackUtil = require('../util/webpack'),
+    ghPages = require('gulp-gh-pages'),
+    runSequence     = require('run-sequence');
```
AlasdairSwan: Nitpick: spacing

AlasdairSwan:
Just a nitpick. Once tests pass 👍

PR #72, opened 2015-05-17
Fix #68: Simplfying font-face Calls

talbs:
This work tackles https://github.com/edx/ux-pattern-library/issues/68.

- - - 

**Reviewers**
* @marcotuts

PR #334, opened 2016-06-09
Add breadcrumbs pattern

bjacobel:
Refs [TNL-4599](https://openedx.atlassian.net/browse/TNL-4599)

Adds breadcrumb styling and pattern example. Uses JS components added to the UITK in [#72](https://github.com/edx/edx-ui-toolkit/pull/72).

http://ux-test.edx.org/bjacobel/breadcrumbs/

## Testing Checklist
- [x] Manually test responsive behavior.
- [ ] Manually test right-to-left behavior.

## Non-testing Checklist
- [x] Consider any documentation your change might need, and which users will be affected by this change.

## Post-review

## Reviewers
- [x] @andy-armstrong 
- [x] @AlasdairSwan 
- [x] @clrux 
- [ ] @chris-mike 

## TODO before merge
- [x] Merge UITK PR first
- [x] release UITK to NPM
- [x] update the UXPL `package.json` git uri to point to npm UITK v 1.2.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "browser-sync": "*",
     "css-loader": "~0.23.1",
     "del": "*",
-    "edx-ui-toolkit": "~0.10.0",
+    "edx-ui-toolkit": "file:../edx-ui-toolkit",
     "extract-text-webpack-plugin": "~1.0.1",
     "file-loader": "0.8.5",
     "font-awesome-loader": "0.0.1",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "browser-sync": "*",
     "css-loader": "~0.23.1",
     "del": "*",
-    "edx-ui-toolkit": "file:../edx-ui-toolkit",
+    "edx-ui-toolkit": "~1.4.0",
     "extract-text-webpack-plugin": "~1.0.1",
     "file-loader": "0.8.5",
     "font-awesome-loader": "0.0.1",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "browser-sync": "*",
     "css-loader": "~0.23.1",
     "del": "*",
-    "edx-ui-toolkit": "~1.4.0",
+    "edx-ui-toolkit": "~1.4.1",
     "extract-text-webpack-plugin": "~1.0.1",
     "file-loader": "0.8.5",
     "font-awesome-loader": "0.0.1",
--- a/pattern-library/sass/patterns/_breadcrumbs.scss
+++ b/pattern-library/sass/patterns/_breadcrumbs.scss
@@ -4,12 +4,12 @@
     line-height: line-height(small);
 
     .nav-item {
-        @include margin-right($baseline/2);
+        @include margin-left($baseline/4);
         display: inline-block;
     }
 
     .fa.fa-angle-right {
-        @include margin-right($baseline/4);
+        @include margin-left($baseline/4);
         display: inline-block;
         color: palette(grayscale, light);
 
--- a/pattern-library/sass/patterns/_breadcrumbs.scss
+++ b/pattern-library/sass/patterns/_breadcrumbs.scss
@@ -8,7 +8,7 @@
         display: inline-block;
     }
 
-    .fa.fa-angle-right {
+    .fa-angle-right {
         @include margin-left($baseline/4);
         display: inline-block;
         color: palette(grayscale, light);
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "css-loader": "~0.23.1",
     "browser-sync": "*",
     "del": "*",
-    "edx-ui-toolkit": "~0.10.0",
+    "edx-ui-toolkit": "git://github.com/edx/edx-ui-toolkit#bf34f8e6b0028b90c07536bd7279325ed6c3fa86",
```
andy-armstrong: Note: don't merge until this can be changed to an actual release number.

pldoc/_components/breadcrumbs.md
```diff
@@ -0,0 +1,30 @@
...
+
+url_styles:             "patterns/_breadcrumbs.scss"
+
+description:            Breadcrumb navigation.
+
+info:                   Accessible tiered navigation using breadcrumbs.
+ui-toolkit-component:   breadcrumbs
+js:                     "/examples/breadcrumbs-js.html"
+---
+<div class="breadcrumbs-example">
```
andy-armstrong: Can you add a pre-rendered version too (like @AlasdairSwan does with the dropdown menu). I think it will be useful to be able to render breadcrumbs server-side sometimes.

andy-armstrong:
👍  Awesome stuff, @bjacobel. If it is straightforward I'd like to see a server-side rendered example, but if you prefer that can come in a follow-on PR.

clrux:
The last page in the breadcrumb shouldn't be a link. Maybe include a heading (it can be `sr-only`) for even more context.

[The W3 breadcrumb "spec"](https://www.w3.org/TR/WCAG20-TECHS/G65)
[An example from the W3](https://www.w3.org/WAI/WCAG20/Techniques/working-examples/G65/ex3.html)

clrux:
Oh, also, if you use a list for the breadcrumb links, screenreaders announce the number of items, which could add additional clarity to how "deep" in they are.

bjacobel:
@clrux Updated with some accessibility improvements.

clrux:
@bjacobel Sorry man, after doing some additional testing, the `nav` only has one child (`ul`) and so doesn't properly announce the correct number of children. It'll always be one.

Check out this example, which I tested, and does work. https://codepen.io/jonneal/pen/ianKu

Turns out we don't really need a list if we're using the `nav` element.

bjacobel:
@clrux no worries! I've updated to remove the `ul` / `li` structure.

bjacobel:
Example of pre-rendered breadcrumbs is live on doc site @andy-armstrong.

Checking Chris off for review as he's already done the a11y review when the components went into the toolkit.

andy-armstrong:
👍  looks awesome. Thanks for the pre-rendered version. 

I recommend having @AlasdairSwan do the second review instead of @dianakhuang (Alasdair, is that okay with you?)

pattern-library/sass/patterns/_breadcrumbs.scss
```diff
@@ -0,0 +1,20 @@
...
+    font-family: $font-family-sans-serif;
+    font-size: font-size(small);
+    line-height: line-height(small);
+
+    .nav-item {
+        @include margin-left($baseline/4);
+        display: inline-block;
+    }
+
+    .fa.fa-angle-right {
```
AlasdairSwan: Does this need both selectors?

pldoc/_components/breadcrumbs.md
```diff
@@ -0,0 +1,54 @@
...
+<div class="breadcrumbs-basic example-set">
+  <h2>Basic breadcrumbs</h2>
+  <header class="has-breadcrumbs hd-3"></header>
+  <a class="add-breadcrumb" href="">Add breadcrumb</a>
+</div>
+
+<div class="breadcrumbs-prerendered example-set">
+  <h2>Pre-rendered breadcrumbs</h2>
+  <header class="has-breadcrumbs hd-3">
+    <div class="sr-is-focusable" tabindex="-1"></div>
```
AlasdairSwan: What is this element used for?

pattern-library/sass/patterns/_breadcrumbs.scss
bjacobel: Probably doesn't need the `.fa` selector, I will remove.

pldoc/_components/breadcrumbs.md
bjacobel: I actually have no idea - this is direct from the Teams implementation of breadcrumbs.

Just asked Chris, he says it's not necessary, so I'll remove it. (Requires doing a UITK patch release, unfortunately.)

bjacobel:
@AlasdairSwan Pushed an update that bumps the UITK dependency to 1.4.1, removing that unnecessary `sr-is-focusable`.

AlasdairSwan:
Looks good @bjacobel 👍

PR #339, opened 2016-06-30
Fix icon class to use font awesome on Alerts and Form sample pages

alisan617:
## Description

- Update icon class to font awesome on Alerts sample page.

- Update icon class to font awesome for the search icon on Form sample page.

[Preview](http://ux-test.edx.org/alisan/alert-icons-missing-in-pattern-library/components/alerts/)

## Post-review

## Reviewers
- [ ] @andy-armstrong 

### DIFF
--- a/pldoc/404.html
+++ b/pldoc/404.html
@@ -6,7 +6,7 @@ permalink: 404.html
 
 <div class="message message-404">
     <div class="alert alert-error" role="alert" aria-labelledby="alert-title-error" tabindex="-1">
-        <span class="icon alert-icon icon-warning" aria-hidden="true"></span>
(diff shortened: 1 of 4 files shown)


andy-armstrong:
@alisan617 I see a couple more icons if I do a search of the repo:

https://github.com/edx/ux-pattern-library/blob/24173366bbb3cec4ad4227f0d7089fa10f8b53e5/pldoc/404.html#L9

https://github.com/edx/ux-pattern-library/blob/24173366bbb3cec4ad4227f0d7089fa10f8b53e5/pldoc/static/js/size-slider.js#L17

👍  with those fixed.

andy-armstrong:
@alisan617 Actually, the second one if probably not a bug, but the name is confusing because it looks like the old scheme.

PR #344, opened 2016-07-21
Make entire repo pass ESLint

bjacobel:
## Description
Previously we were linting only a small subset of "source" files and not including some directories that contained a lot of JS code (notably `./gulp`). This lints the whole repo, and fixes all the existing lint violations. 

(I went ahead and did this because my editor has no way to turn off linting on a per-file basis, and all the red in the gulp scripts was making it hard to concentrate while debugging other issues.)

## Reviewers
- [x] @andy-armstrong 
- [x] @alisan617 

### DIFF
--- a/gulp/tasks/browserSync.js
+++ b/gulp/tasks/browserSync.js
@@ -1,6 +1,8 @@
-var gulp            = require('gulp'),
-    browserSync     = require('browser-sync'),
-    config          = require('../config').browserSync;
+'use strict';
+
+var gulp = require('gulp'),
+    browserSync = require('browser-sync'),
+    config = require('../config').browserSync;
 
 gulp.task('browserSync', ['jekyll-build'], function() {
     browserSync(config);
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,5 @@
 _preview_site
 _site
 node_modules
+!pldoc/_includes/examples
+pldoc/public
--- a/gulp/.eslintrc.json
+++ b/gulp/.eslintrc.json
@@ -1,5 +1,6 @@
 {
     "rules": {
-        "no-console": "off"
+        "no-console": "off",
+        "strict": ["error", "global"]
     }
 }
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -1,102 +1,104 @@
 // gulp pattern library + site configuration
 
+'use strict';
+
     // pattern library
-var patternLibrarySrc           = './pattern-library',
-    local                       = './_site',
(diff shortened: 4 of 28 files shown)


gulp/tasks/browserSync.js
```diff
@@ -1,7 +1,11 @@
...
-    browserSync     = require('browser-sync'),
-    config          = require('../config').browserSync;
+var gulp = require('gulp'),
+    browserSync = require('browser-sync'),
+    config = require('../config').browserSync;
 
-gulp.task('browserSync', ['jekyll-build'], function() {
-    browserSync(config);
-});
+(function() {
```
andy-armstrong: Nit: I think if you're moving the variables out you should probably remove the immediately invoked functions too. My understanding is that they are not necessary in node since it wraps each module inside a function itself.

andy-armstrong: I see now that you're adding them. Is that to make ESLint happy? Do we need to define a rule override for the gulp directory?

andy-armstrong:
👍  Looks good to me, @bjacobel. Feel free to implement or ignore my nit as you prefer.

alisan617:
It would be great if this is a task. When new gulp js files are edited, lint all. By default, lint all is off, as it ends up taking longer to lint all.

bjacobel:
@alisan617 Yes, I agree, having the Gulp tasks and all linted when you save a source file while the watcher is running isn't ideal. How about a new task, `lint-src`, that lints only our "source" files, and that's what get called when gulp watch is triggered?

gulp/tasks/browserSync.js
bjacobel: Yep, you're right. IIFEs don't do anything extra in node as each module has its own scope.

I added the IIFEs to make eslint's `strict` rule happy (`'use strict';` must be invoked at function scope) but it looks safe to turn that rule off for the `/gulp` directory as the files in it are commonjs modules.

alisan617:
@bjacobel Yes. That is a good idea.

bjacobel:
@andy-armstrong @alisan617 Ready for re-review.

andy-armstrong:
@bjacobel 👍 once you fix the build issue. Thanks for taking the time to address my nit.

bjacobel:
Nice catch from Travis - `gulp-eslint` > 3.0 (which brings in eslint 3.1.1)  will not import `eslint-config-airbnb-legacy/base` due to the slash in the name, I think. Something to be aware of when we do https://github.com/edx/eslint-config-edx/issues/16.

PR #345, opened 2016-07-22
Updating color palettes and color swatch layout

clrux:
# [UXPL-36](https://openedx.atlassian.net/browse/UXPL-36)

## Summary of work

This work updates our color palettes to provide primary and secondary values that work on both white and our light gray.

## Additional details and background

Previously, our core blue and core pink did not work on gray and in some cases, they barely worked on white. As developers begin to utilize the pattern library and our colors more heavily, we needed to provide options that won't cause accessibility failures in tests and that allow for an easy, consistent approach to development.

Along with updated colors, with the help of the UX team, I've simplified the number of options. Our primary color palettes was reduced from 10 to 5, with the base options being accessible on both white and gray. I've eliminated our "information" color altogether, per UX request. I've also reduced our grayscale sets to just one, basing the gray off of Gabe's gray and eliminating our "cool" gray.

I've updated the color swatch layout to better showcase these color options but displaying them on various background colors. This not only gives developers and sense of what any color might look like in context with a background color, but it also allows our accessibility tool to check the contrast with any combination, making the overall process more robust.

Each color swatch is a button making the interface accessible to keyboard-only users, and each button is clickable revealing more information about the color, recommended use, and developer variables. The information was built using a JSON file so is easily maintainable.

The goal of the pattern library is "easy consistency". Too many color options reduces consistency, while too few options don't provide enough variation. The UX team and I felt that 5 color options, with the base option accessible, provides a nice balance between the two.

I've also been working with Gabe from marketing to sync up the print-based colors with these, and visa versa.

---

I went through and updated color variables use in other patterns to use the new variables. In the event where a variable was removed, I am using the closest option. As a result, the overall layout and presentation changed, some ways more subtle than others.

---

Before bumping the version in platform, I need to update any instances in platform that utilize stale variable names that were removed during this process.

## Preview sandbox

Soon...

## Reviewers

- [x] @chris-mike or @roderickmorales
- [x] @andy-armstrong or @bjacobel (@bjacobel :thumbsup:)

### For your information only

@dsjen, @cptvitamin, @cahrens, @alisan617, @AlasdairSwan

### DIFF
--- a/pattern-library/sass/global/_settings.scss
+++ b/pattern-library/sass/global/_settings.scss
@@ -54,114 +54,71 @@ $bp-screen-xl:                  1280px !default;
 // color: config
 $transparent:                   rgba(0, 0, 0, 0);
 
-$black:                         rgb(17, 16, 16);
-$black-t:                       rgb(0, 0, 0); // true black
-$white:                         rgb(252, 252, 252);
-$white-t:                       rgb(255, 255, 255); // true white
-$gray:                          rgb(167, 164, 164);
-$gray-cool:                     rgb(158, 177, 185);
+$black:                         rgb(22, 22, 22);
+$white:                         rgb(255, 255, 255);
+$gray:                          rgb(138, 143, 143);
 $red:                           rgb(178, 6, 16);
-$yellow:                        rgb(253, 188, 86);
-$green:                         rgb(37, 184, 90);
-$blue:                          rgb(0, 121, 188);
-$pink:                          rgb(203, 89, 141);
+$yellow:                        rgb(255, 192, 31);
+$green:                         rgb(0, 129, 0);
+$blue:                          rgba(0, 117, 180, 100);
+$pink:                          rgba(194, 56, 125, 100);
 
 // colors:                      brand
 $brand-primary:                 $blue !default;
-$brand-primary-accent:          rgb(14, 166, 236) !default;
 $brand-secondary:               $pink !default;
-$brand-secondary-accent:        rgb(242, 108, 170) !default;
 
 // colors: utilities
-$success: $green !default;
-$warning: $yellow !default;
-$error: $red !default;
-$information: mix($gray-cool, $brand-primary, 70%) !default;
+$success:                       $green !default;
+$warning:                       $yellow !default;
+$error:                         $red !default;
+$information:                   $blue !default;
 
 // colors: palettes
 $palettes: (
     primary: (
-        base:       $brand-primary,
-        light:      mix($white, $brand-primary, 33%),
-        x-light:    mix($white, $brand-primary, 66%),
-        dark:       mix($black, $brand-primary, 33%),
-        x-dark:     mix($black, $brand-primary, 66%),
-        accent:     $brand-primary-accent,
-        x-accent:   saturate($brand-primary-accent, 25%),
-        trans:      transparentize($brand-primary, 0.50),
-        x-trans:    transparentize($brand-primary, 0.75),
-        xx-trans:   transparentize($brand-primary, 0.95)
+        dark:       rgba(6, 86, 131, 100),
+        base:       rgba(0, 117, 180, 100),
+        accent:     rgba(41, 145, 195, 100),
+        back:       rgba(204, 227, 240, 100),
+        x-back:     rgba(242, 248, 251, 100)
     ),
     secondary: (
-        base:       $brand-secondary,
-        light:      mix($white, $brand-secondary, 33%),
-        x-light:    mix($white, $brand-secondary, 66%),
-        dark:       mix($black, $brand-secondary, 33%),
-        x-dark:     mix($black, $brand-secondary, 66%),
-        accent:     $brand-secondary-accent
+        dark:       rgba(152, 44, 98, 100),
+        base:       rgba(194, 56, 125, 100),
+        accent:     rgba(224, 155, 190, 100),
+        back:       rgba(243, 215, 229, 100),
+        x-back:     rgba(252, 245, 248, 100)
     ),
     grayscale: (
-        base:       $gray,
-        light:      mix($white, $gray, 50%),
-        x-light:    mix($white, $gray, 75%),
-        xx-light:   mix($white, $gray, 85%),
-        white:      $white,
-        white-t:    $white-t,
-        dark:       mix($black, $gray, 40%),
-        x-dark:     mix($black, $gray, 60%),
-        black:      $black,
-        black-t:    $black-t,
-        trans:      transparentize($gray, 0.50),
-        x-trans:    transparentize($gray, 0.75),
-        xx-trans:   transparentize($gray, 0.95)
-    ),
-    grayscale-cool: (
-        base:       $gray-cool,
-        light:      mix($white, $gray-cool, 50%),
-        x-light:    mix($white, $gray-cool, 75%),
-        xx-light:   mix($white, $gray-cool, 85%),
-        dark:       mix($black, $gray-cool, 50%),
-        x-dark:     mix($black, $gray-cool, 75%),
-        trans:      transparentize($gray-cool, 0.50),
-        x-trans:    transparentize($gray-cool, 0.75),
-        xx-trans:   transparentize($gray-cool, 0.95)
+        dark:       rgba(17, 17, 17, 100),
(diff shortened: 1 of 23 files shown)


pattern-library/sass/global/_settings.scss
```diff
@@ -54,114 +54,71 @@ $bp-screen-xl:                  1280px !default;
...
-$brand-primary-accent:          rgb(14, 166, 236) !default;
 $brand-secondary:               $pink !default;
-$brand-secondary-accent:        rgb(242, 108, 170) !default;
 
 // colors: utilities
-$success: $green !default;
-$warning: $yellow !default;
-$error: $red !default;
-$information: mix($gray-cool, $brand-primary, 70%) !default;
+$success:                       $green !default;
```
bjacobel: Personally I don't love the style of indenting for table-like formatting, but it is consistent with the rest of the file, so ¯\_(ツ)_/¯

pldoc/static/js/color-contrast.js
```diff
@@ -65,6 +52,20 @@ define([
             });
         },
 
+        rgbaToHex: function(rgba) {
+            var rgbaValue = rgba.match(/^rgba?[\s+]?\([\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?/i),
```
bjacobel: Yikes 😬 Maybe add a comment showing what kind of vale this will match? I'm assuming just `rgba(x, x, x, x)`

pldoc/static/js/colors.js
```diff
@@ -0,0 +1,132 @@
+define([
+    'jquery'
+], function($) {
+    'use strict';
+
+    var $pane = $('#colors-preview .info-pane'),
+        $title = $($pane).find('.color-info-title'),
```
bjacobel: Shouldn't need to wrap `$pane` in a call to `$()` if it's already a jQuery object.

pldoc/static/js/colors.js
```diff
@@ -0,0 +1,132 @@
...
+            ColorHandler.listenForKeypress();
+            ColorHandler.listenForClose();
+        },
+
+        setupJSONColors: function() {
+            $.ajax({
+                async: false,
+                url: '/public/js/colors.json',
+                complete: function(data) {
+                    colors = data.responseJSON;
```
bjacobel: Couple things here:

- You're doing an ajax request for a local file, which seems strange, have you looked into just requiring the json with requireJS? I think that's possible, but I could be wrong.
- You don't do anything with `colors` after you get the JSON response?
- Synchronous ajax will lock the browser while this file is fetched, IIRC.

pldoc/static/js/colors.js
```diff
@@ -0,0 +1,132 @@
...
+                }
+            });
+        },
+
+        listenForClick: function() {
+            var classes;
+
+            $('#colors-preview').on('click', 'button.swatch', function() {
+                $button = $(this);
+                classes = $($button).attr('class').split(/\s+/);
```
bjacobel: Ditto here what I said with `$($pane)` above - `$button.attr` should work fine without the extra jQuery call.

pldoc/static/js/colors.js
```diff
@@ -0,0 +1,132 @@
...
+            });
+        },
+
+        listenForKeypress: function() {
+            var key;
+
+            $(document).on('keydown', function(e) {
+                key = e.which;
+
+                if (key === 27) {
```
bjacobel: Don't like magic numbers - can we either make this a variable at the top of the file so as to document which key it is, or (better) pull it in from some constants file somewhere? Don't know if we have one of those already.

pldoc/static/js/colors.js
```diff
@@ -0,0 +1,132 @@
...
+            $button.focus();
+        },
+
+        clearValues: function() {
+            $title.text('');
+            $desc.text('');
+            $back.html('');
+        },
+
+        rgbaToHex: function(rgba) {
```
bjacobel: This looks duplicated from where I commented on it above, can it be pulled out into someplace common?

pldoc/static/js/colors.json
```diff
@@ -0,0 +1,174 @@
+[{
```
bjacobel: If using requireJS to import this json file doesn't work, it should also work to convert this into a `*.js` file that just does:

```javascript
module.exports = {
  your: 'json',
  goes: 'here'
};
```

and I can't see any reason requirejs wouldn't handle that fine. As a bonus, then it would get bundled into all the other Javascript, and you wouldn't need to add the Gulp tasks for JSON files you added above.

bjacobel: Also, this .json file isn't valid JSON - it's an array. JSON has to be an object at the top level. You could do:

```javascript
{
  colors: [
    <contents of this file>
  ]
}
```

and then it'd become valid JSON.

bjacobel:
Looks pretty good so far - I think you can simplify things a bit by including that new json file with requirejs (and even further by just making it a `.js` file that exports an object). Let me know when you're ready for re-review.

Also I'm working on fixing the reason this isn't on a preview site - I broke the Gulp webpack scripts :grimace:

pldoc/static/js/colors.js
clrux: I need the synchronosity here, as we need to run functions on each color swatch, but only after it's ready. I'll look into including a JSON file as RJS.

clrux: You can require a JSON file, but only with a [RequireJS Plugin](https://github.com/millermedeiros/requirejs-plugins).

pldoc/static/js/colors.js
clrux: We have constants in the UI Toolkit (utils), but those aren't available here. FWIW, 27 won't change. That's always the number for Escape. Not sure if that matters.

pldoc/static/js/colors.js
clrux: I tried including the `AccessibilityColorContrast` from `color-contrast.js` but that file isn't setup to return what I need so it doesn't work, and needs to be duplicated. Unfortunately I don't have time to re-write that file as this needs to be done by like Friday if not sooner.

FWIW, this isn't part of the pattern library that's available for folks to use, so I feel like we have a bit of flexibility here. This is exclusive to the preview site.

pldoc/static/js/colors.json
bjacobel: Update: I am dumb and wrong, `[]` is perfectly valid JSON.

clrux:
@bjacobel Thanks for the feedback. Review once more please?

gulp/config.js
```diff
@@ -78,6 +78,7 @@ module.exports = {
         // documentation site
         pldoc_src: pldocSrc + '/static/js',
         pldoc_src_files: pldocSrc + '/static/js/**/*.js',
+        pldoc_src_json: pldocSrc + '/static/js/**/*.json',
```
bjacobel: Now that your colors file is `js` and brought in through requirejs - it shouldn't need any special treatment - you can remove this and the new Gulp task below

pldoc/static/js/colors.js
bjacobel: I was actually just talking about this `rgbaToHex` function - it's been added to `color-contrast.js` as well. Can you add it to a utils file, and import it and use it both there and here to remove duplication?

pldoc/static/js/colors.js
bjacobel: I'd just add a comment specifying what key 27 is for clarity, then.

AlasdairSwan: @clrux the UI Toolkit has constants.js that maps keyCodes and can be used here (https://github.com/edx/edx-ui-toolkit/blob/master/src/js/utils/constants.js)

clrux: Is this available in the pattern library?

bjacobel: yeah, you should be able to get it with `'edx-ui-toolkit/js/utils/constants.js'` in requirejs

clrux:
@bjacobel Another?

bjacobel:
LGTM 👍

bjacobel:
(Make sure to rebase, though)

PR #351, opened 2016-08-01
Update styling for search box

bjacobel:
## Description

Updates styling of `<button>`s immediately following `<input>`s to remove spacing between them and match styling and sizing.

http://ux-test.edx.org/bjacobel/search-tweaks/components/forms/

(old: [spacing between box and button, rounded corners, mismatched sizes](https://i.bjacobel.com/20160801-zcawc.png))

Checked in FFX, Safari and Chrome and confirmed the styling is consistent.
  
## Testing Checklist
- ~~Manually test right-to-left behavior.~~ N/A

## Post-review

## Reviewers
- [ ] @andy-armstrong 
- [x] @chris-mike 
- [x] @clrux 

### DIFF
--- a/pattern-library/sass/_normalize.scss
+++ b/pattern-library/sass/_normalize.scss
@@ -355,21 +355,9 @@ input[type="number"]::-webkit-outer-spin-button {
  */
 
 input[type="search"] {
-  -webkit-appearance: textfield; /* 1 */
   -moz-box-sizing: content-box;
   -webkit-box-sizing: content-box; /* 2 */
   box-sizing: content-box;
-}
-
-/**
- * Remove inner padding and search cancel button in Safari and Chrome on OS X.
- * Safari (but not Chrome) clips the cancel button when the search input has
- * padding (and `textfield` appearance).
- */
-
-input[type="search"]::-webkit-search-cancel-button,
-input[type="search"]::-webkit-search-decoration {
-  -webkit-appearance: none;
 }
 
 /**
--- a/pattern-library/sass/patterns/_forms.scss
+++ b/pattern-library/sass/patterns/_forms.scss
@@ -257,6 +257,15 @@ $input-width: (
             border-color: palette(error, text);
         }
     }
+
+    // Input fields immediately followed by a button (ie, search boxes) should eliminate the spacing
+    // between them and adopt matching styling/sizing to emphasize that they are connected.
+    &[type=search] + button {
+        border-top-left-radius: 0;
+        border-bottom-left-radius: 0;
+        font-size: font-size(base);
+        margin-left: (-1 * $baseline / 4);
+    }
 }
 
 
--- a/pldoc/_components/forms.md
+++ b/pldoc/_components/forms.md
@@ -111,8 +111,9 @@ info:               Forms allow users to interact with the interface, set prefer
                 <label class="field-label" for="search-03">Search (with visible label and icon)</label>
                 <input class="field-input input-text" type="search" id="search-03" name="search-03" placeholder="Search this website">
                 <button class="btn-brand btn-small" type="button">
-                    <span class="icon fa fa-search" aria-hidden="true"></span>
-                    <span class="sr-only">Search</span>
+                    <span class="icon fa fa-search" aria-hidden="true">
+                        <span class="sr-only">Search</span>
+                    </span>
                 </button>
             </div>
         </fieldset>


clrux:
Awesome! What do you think about rounding the top-right/bottom-right corners to match our other buttons a little closer?

bjacobel:
@chris-mike's position was I think that all four corners should be square (see attached screenshot on [TNL-4726](https://openedx.atlassian.net/browse/TNL-4726)), but it's easy to implement either way.

andy-armstrong:
@bjacobel It looks like the icon doesn't center exactly in the box. There is a right margin that is unnecessary, presumably for the situation where the label is followed by some text.

pattern-library/sass/patterns/_forms.scss
```diff
@@ -257,6 +257,14 @@ $input-width: (
             border-color: palette(error, text);
         }
     }
+
+    // Input fields immediately followed by a button should eliminate the spacing between them
```
andy-armstrong: I'm not sure, but I wonder if this will be too general a rule. Having said that, I'm trying to think of a situation where you might need a button afterwards that is not associated with the input, but I haven't thought of one. It is just that I was expecting this rule to be associated with a class like ``search-form``.

andy-armstrong:
@bjacobel This looks great. 👍 

For the record, I'm in favor of the square corners as I think the button will look odd with curves only on the right. However, I'd be interested to see what it looks like with round corners.

BTW, you are going to run into a problem using this code in edx-platform. @clrux's color changes are incompatible, so you (or he) will have to resolve them first. :-(

clrux:
I like the square corners too (dislike rounded things, personally); it was more of a question. I'm OK with this 👍 

As for the color updates, I'm afraid with less than two months until our deadline I don't have time. I recon a find/replace may work alright? But yes, be sure and check out the new color variable names - they've been slimmed down quite a bit and marker names were updated to better reflect use.

bjacobel:
Confused about what you mean, this doesn't modify any colors?

bjacobel:
FYI, here it is with rounded corners only on the outside edge.



It looks pretty weird when the button wraps to the next line on small viewports:

andy-armstrong:
@bjacobel What I meant is that if you cut a new Pattern Library release it will include your changes and Chris's. If you then use it in edx-platform, there will be Sass compilation errors trying to use colors that no longer exist.

pattern-library/sass/patterns/_forms.scss
bjacobel: Mmm. Good thought. We don't currently have such a class - I could set it up this way, though:

```css
input[type=search] .form-input {
  & + button {
  }
}
```

so that it only applies to search boxes?

andy-armstrong: That sounds like a good solution, thanks.

andy-armstrong:
@bjacobel I'm going to make a PR to update the platform to the latest Pattern Library release, and I'll then fix any color errors that arise.

andy-armstrong:
@bjacobel @clrux Here's the new ticket for cleaning up the platform colors: https://openedx.atlassian.net/browse/TNL-5135

chris-mike:
The button wrap / break is unfortunate.  I assume the text entry field would be adjustable, so that this is not likely to break on the smallest views. 

I like the hybrid - square form, rounded edge. Just one opinion. Happy to stick with all square if others feel strongly about it. I do feel strongly that the form field (left-side) should be a square).

One final thought - do we need something to easily delete the content? If that's for another point in time, that works, too (esp. since it doesn't exist today).

Looks good to me.

👍

bjacobel:
Updated code and preview site with: 
- hybrid corners (square on left, round on right)
- HTML5 button to clear box (no JS needed)
- fixes for misaligned icons (this was actually an issue with the markup, the padding is removed when the `.sr-only` element is placed inside the icon where it should be)

pattern-library/sass/_normalize.scss
```diff
@@ -355,24 +355,12 @@ input[type="number"]::-webkit-outer-spin-button {
...
 
 input[type="search"] {
-  -webkit-appearance: textfield; /* 1 */
   -moz-box-sizing: content-box;
   -webkit-box-sizing: content-box; /* 2 */
   box-sizing: content-box;
 }
 
 /**
- * Remove inner padding and search cancel button in Safari and Chrome on OS X.
```
andy-armstrong: Wow, so that's why we didn't get the cancel button! I wonder why this was deliberately disabled...

bjacobel: It's not consistent between browsers (FF and IE don't do it), so maybe somebody was thinking we'd build our own that worked everywhere and would turn the native one off so they didn't conflict?

pattern-library/sass/patterns/_forms.scss
andy-armstrong: What's your plan with this approach? I would prefer it to removing padding before all buttons that happen to be next to an input. Did it not work out?

pattern-library/sass/_normalize.scss
andy-armstrong: That's a reasonable theory. I think the way you have it is better, because folks who are used to Chrome or Safari would expect to see the cancel button. I can see that we might want to implement the same behavior into Firefox and IE at some point, but it doesn't seem very high priority.

pattern-library/sass/patterns/_forms.scss
bjacobel: Oh - just forgot. Pushing that change up now.

bjacobel:
@andy-armstrong should I hold off on merging/releasing this until https://github.com/edx/edx-platform/pull/13144 lands?

andy-armstrong:
@bjacobel Feel free to merge this and cut a new release, and then I'll update https://github.com/edx/edx-platform/pull/13144 to use your version.

PR #353, opened 2016-08-03
Avoid underlining all anchor elements

rlucioni:
@andy-armstrong @clrux I think this would prevent all anchor elements on pattern library pages from being underlined on hover. I'm not sure why this was introduced.

### DIFF
--- a/pattern-library/sass/global/_helpers.scss
+++ b/pattern-library/sass/global/_helpers.scss
@@ -133,7 +133,6 @@
     &:active,
     &:focus {
         color: $link-focus-color;
-        text-decoration: underline;
     }
 
     // STATE: is disabled


clrux:
@rlucioni I added this recently. It's good practice to use two forms of visual change. For example, a color shift and an underline. This is in case the color change is imperceptible (we sometimes use a lighter shade of our blue, but just barely) or for users who may be color blind. Using color alone isn't sufficient.

andy-armstrong:
@rlucioni Thanks for tracking this down before I even got to sit down to investigate!

@clrux It seems confusing that we are now inconsistent about how we show buttons. If they are links styled to look like buttons they get the underline, but if they are true buttons they don't. You can see this on this page in the "Links with visual styling applied" section:

http://ux.edx.org/components/buttons/

Those buttons, and only those, have underlines on hover. Was this just an oversight that we can fix locally, or did you intend button-styled links to have this behavior? It looks very odd in Programs where some buttons have underlines and some don't.

Maybe the answer is that for a11y reasons we should be switching all of these links over to ``<button>`` tags anyway. However, in the interim I think we need to remove the underlines from button-styled links.

clrux:
@andy-armstrong Well, those are links that look like buttons. But we can apply different styling to those buttons/links only, if that's preferred. I do agree that it looks a little funky - this was an oversight on my part.

I will try to get this quick update in tomorrow - and in before Friday. Unless you or Brian are working on something in there now and can roll it in?

rlucioni:
I'm closing this under the assumption that @andy-armstrong will be removing underlines from links styled to look like buttons separately.

PR #363, opened 2016-08-29
Adaptive placeholders

bjacobel:
## Description

Created as part of [a discussions ticket](https://openedx.atlassian.net/browse/TNL-5166) to remove `placeholder` attributes due to their accessibility issues, I created a new "adaptive" placeholder that provides a persistent visible label for form inputs without requiring (much) extra room in the design. Inspired by Target, Dollar Shave Club, CircleCI and others.

Features:
- No Javascript!
- Super accessible! (I think?)
- Kind of fun to use!
- Decently backwards-compatible (just add a wrapping div and one new attribute to existing `input`s)
- Good browser support
- Should seamlessly work with inputs of all heights and font sizes

Anti-features:
- Doesn't work with any input where we're already using HTML5 validation API to conditionally display CSS
- Requires input to be `required` (i.e., won't work for optional inputs, unless you specify `novalidate` on the form)

[Demo/sandbox](http://ux-test.edx.org/bjacobel/better-placeholders/components/forms/)
## Testing Checklist
- [ ] Manually test right-to-left behavior. (known issue with button on search input, fixing in separate PR)
- [ ] Manually test a11y support.
## Non-testing Checklist
## Post-review
## Reviewers
- [ ] @andy-armstrong 
- [ ] @chris-mike or other from @edx/edx-ux-team 
- [ ] @clrux or other from @edx/edx-accessibility 

### DIFF
--- a/pattern-library/sass/patterns/_forms.scss
+++ b/pattern-library/sass/patterns/_forms.scss
@@ -293,13 +293,17 @@ $input-width: (
 
     .field-input:focus,
     .field-input:valid {
-        & + .field-hint {
+        & ~ .field-hint {
             background-color: $white;
             font-size: font-size(small) !important;
             margin-top: (-$baseline / 2 - 2);
             @include margin-left(rem($baseline / 2));
             padding: 2px;
         }
+    }
+
+    .input-alt ~ .field-hint {
+        padding: spacing-vertical(small) 0;
     }
 }
 
--- a/pattern-library/sass/patterns/_forms.scss
+++ b/pattern-library/sass/patterns/_forms.scss
@@ -268,6 +268,31 @@ $input-width: (
     }
 }
 
+.field-hint.placeholder {
+    padding: rem($baseline/2);
+    margin: 0;
+    position: absolute;
+    border: 1px solid transparent;
+    width: auto;
+    font-size: inherit !important;
+    color: palette(grayscale, base);
+    background-color: transparent;
+    pointer-events: none;
+    transition: margin 0.2s ease, font-size 0.2s ease, background-color 0.2s ease;
+
+    * {
+        margin: 0;
+        padding: 0;
+    }
+
+    &.activated {
+        background-color: $white;
+        font-size: font-size(small) !important;
+        margin-top: (-$baseline / 2 - 2);
+        @include margin-left(rem($baseline / 2));
+        padding: 2px;
+    }
+}
 
 // ----------------------------
 // #SELECT
--- a/pattern-library/sass/patterns/_forms.scss
+++ b/pattern-library/sass/patterns/_forms.scss
@@ -268,29 +268,38 @@ $input-width: (
     }
 }
 
-.field-hint.placeholder {
-    padding: rem($baseline/2);
-    margin: 0;
-    position: absolute;
-    border: 1px solid transparent;
-    width: auto;
-    font-size: inherit !important;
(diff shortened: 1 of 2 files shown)


roderickmorales:
@bjacobel I like this approach and was hopping to use it in our new publisher tool. My only feedback is that I'd like to take a pass at the colors/styles but we can address those once this gets merged.

chris-mike:
While I do like the look and the overall interaction, here are my two reactions.

1. I'd like to see text alignment be a little more consistent. Clicking into the form fields, my cursor sometimes goes to the left-most point (basically, no padding). I think padding should be added.  There is also some padding concern vertically. I wonder if this means we have to have things at a min height (that is bigger than what we're currently considering in terms of page layout).   (see screen shot of my experience using "auto fill"). 
screen shot 2016-08-30 at 11 46 32 am

2. The other question I have is whether we need to reserve the top of a field for the actual "question" rather than the guiding instructional cue.  I talked with someone about this recently (do not remember who) about taking the general approach that question text is on top, guidance/help text is below and at a smaller font size.  I'm not advocating for one or the other. Sometimes these are just standard form fields... other times they are cues for how we might evolve capa fields, or add a post entries in the discussion forum.  That having been said, we've recently started building guidance (slightly smaller font) to questions in CAPA (for accessibility).

bjacobel:
1. Hoo boy, I hadn't considered the Chrome autofill stuff. That doesn't look great. I'll think about how to solve that.

2. I agree - this change requires re-thinking how we label fields. I don't think with this change examples should serve as the placeholder, and would rather see for example "Your Name" as the text in the adaptive placeholder that moves (with an example below, if we really need it?) But I'm not sure what the accessibility implications of that are.

chris-mike:
I assume the second field shows an autofill error.  It's "normal" when I do not do autofill.

bjacobel:
[I was able to repro the autofill issue](https://i.bjacobel.com/20160830-w9ajq.png). Looks like Chrome doesn't provide any hint to the HTML or CSS that it's trying to autofill that form field, so there's no way to detect it and move the text out of the way. I think this pattern might be incompatible with any scenario where there's a chance the user might be using autofill - which is pretty much every scenario besides search. Wow. That sucks.

bjacobel:
It's not clear we'd even be able to avoid autofill issues if we were doing this with Javascript. wtf chrome?!

chris-mike:
It's inelegant, but it's also temporary. Here's the same experience with a different design.
screen shot 2016-08-31 at 11 45 55 am
screen shot 2016-08-31 at 11 46 28 am
screen shot 2016-08-31 at 11 46 36 am

bjacobel:
Huh, so other people are having problems with the autofilling too...

IDK. I don't like adding a pattern that has more edge cases than actual working cases. I'm considering pulling this PR.

roderickmorales:
We discussed briefly the future of how we introduce new patterns into the library and I would vote for pausing this PR until we hash out a process. Thoughts?

bjacobel:
@roderickmorales Yes, I agree - it's a shame because I wanted to use this pattern in Discussions to get around some of our a11y tickets for October, but I can find another way.

roderickmorales:
@bjacobel and I think you hit the nail on the head. In an ideal world we wouldn't start development until all questions are answered. I will encourage you to raise your concerns with @chris-mike and work together to design a solution that will get approved by all parties before it makes it into code development.

PR #367, opened 2016-09-19
Simplify and improve UXPL development/build process

bjacobel:
Simplify and improve UXPL development/build process

## Description
- Sped up asset compilation dramatically. Before we were looking at a minimum of **35 seconds** every time you re-ran `pldoc-styles`, in addition to the time to copy other assets. Now, Webpack can make a production bundle of _all_ the assets (Javascript, fonts, CSS) in around 30 seconds from cold start - but, much better, there is now a development mode that loads up in 19 seconds and caches/recompiles modules based on the contents that are diffed. The new time to recompile a change of the pattern library sass and get it into a new webpack bundle is about **16 seconds**, which we know to be the same speed as running `node-sass` directly on the Pattern Library root, so as we optimize that process (separate ticket to upgrade Susy) that time should drop as well.

- Removed all content from and references to the `demo`. It was something Brian Talbot created that hadn't been used or modified since, and in my opinion was superseded by the pattern library documentation but continued to exist alongside (actually, _inside_) it.

- Streamlined the asset pipeline, bridged gaps and removed dead paths. Issues like [FEDX-216](https://openedx.atlassian.net/browse/FEDX-216), where assets were rebuilt but never sent to the browser, and the issue where pldoc previews would not get any fonts (no ticket but longstanding annoyance) are now fixed. This involved removing a lot of Gulp tasks that didn't do anything, as far as I can tell. The basic Gulp entry point still builds Jekyll and starts Webpack, but some of the intermediate steps that I don't think anybody was calling directly are gone. If I broke your workflow, let me know.

- Removed the step of the build process where we would lint your javascript before rebundling it because it was slowing us down. I think that's the job of an editor/IDE, not an asset pipeline.

- Implemented `webpack-dev-server` to take advantage of hot module reload and Webpack module caching. Now when you save a file, nothing ever needs to touch the filesystem - WDS invokes Webpack, figures out what modules need to be rebuilt based on the dependency graph, triggers a hot module reload on the browser, serves the emitted assets out of memory and proxies all the non-Webpack things through to the Jekyll server.

- As a result of the above change, the doc site now comes up at **localhost:8080** rather than **localhost:4000**. Added some documentation to the server's startup messages about this.

- Removed a ton of stuff from `package.json` - some we weren't using at all and some we didn't need anymore.

- Moved all the Webpack config into `webpack.config.js` - previously it was generated dynamically in one of the gulp tasks. Now it's much more parseable to see what exact settings Webpack is going to get when you start it up, and you can run it from the command line as well.

- The above enabled making it so that you don't need Gulp to run the development site - there's a script in `npm start` that does all the same stuff as `$(npm bin)/gulp`, even a little faster I think.

- Make Travis lint and try to make a production build on every push and PR, using both node 4 (LTS, current support, uses NPM 2) and node 6 (stable, migration planned soon, uses NPM 3)

- Removed `_zappr.yml` - I forgot to remove it when we turned that off.

Basically we're making Webpack handle as much of this asset processing/bundling stuff as possible, because it's great at it and we get speed boosts when we stay inside the ecosystem. Gulp's role is reduced because we don't need two asset managers and Webpack is better at it 😜

## Known issues
- In development, the doc site throws a 404 for `pattern-library-doc.css`. This is because there is no separate CSS file built in development (the CSS is completely bundled inside the Javascript, so it can be reloaded with HMR) but it would be too complex to have two different Jekyll templates - one that requests the CSS file for production, and one that doesn't for development. So in development we request the CSS file anyway, knowing it will fail and that the Webpacked CSS will be there inside the JS bundle to style the page anyway.

- In development, there is a short flash of unstyled content (FOUC) when we navigate from page to page. This is because we have a hybrid application (server-side HTML, but client-side JS/CSS) and the HTML loads before the JS bundle (containing the CSS) is fully unpacked. The benefit to doing it this way is that we get Hot Module Reload in development, which is awesome and faster than BrowserSync. Living with the FOUC also allows us to have shorter asset build times - otherwise we have to use ExtractTextPlugin (which we do for the prod build) which is ~2x as slow at building Sass.

- I haven't yet tested `gulp doc-publish`. I'll do it after I get thumbs here, because it will wipe out the existing doc site.

## Sandbox
http://ux-test.edx.org/bjacobel/fix-css/

## Testing Checklist
- [x] Manually test responsive behavior.
- N/A Manually test right-to-left behavior. (We don't currently build an RTL version of the pattern library documentation, but I have an idea of how we can do it without slowing things down too much using this new webpack architecture).

## Non-testing Checklist

## Reviewers
- [ ] @alisan617
- [x] @andy-armstrong
- [ ] @dsjen
- [ ] @roderickmorales
- [ ] @thallada
- [ ] @AlasdairSwan


gulp/tasks/webpack.js
```diff
@@ -0,0 +1,68 @@
...
+    gulpUtil = require('gulp-util'),
+    webpack = require('webpack'),
+    WebpackDevServer = require('webpack-dev-server'),
+    gitUtils = require('../util/gitUtils'),
+    config = require('../config.js'),
+    wpConfig = require('../../webpack.config.js'),
+    wpConfigFactory = wpConfig.configFactory,
+    wpConfigDevServer = wpConfig.devServerConfig;
+
+/* Builds the pattern library doc into /pldoc/public.
```
andy-armstrong: Nit: I've never seen multi-line comments like this in our codebase. We either use ``//`` or do a JSDoc-style comment starting with ``/**``:

http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/javascript_guidelines.html#documentation

Not a big deal if you prefer doing it like this, but just fyi.

andy-armstrong: I would probably make these JSDoc-like, with a single line summary as the first line. They aren't defining functions per se, but they are defining conceptual blocks.

roderickmorales:
@bjacobel my only question is around the "demo" directory. Was that an attempt on Brian to have static samples of our patterns to match production? I think something like that will be valuable to have but maybe we can discuss how to set it up, so it makes sense to everyone.

👍

package.json
```diff
@@ -22,59 +22,43 @@
...
     "webpack": "~1.12.14",
     "webpack-stream": "~3.2.0"
   },
   "scripts": {
-    "production": "gulp build-production",
     "gulp": "gulp",
-    "development": "gulp build-development",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "start": "rm -rf pldoc/public && jekyll serve --incremental & webpack-dev-server && fg"
```
dsjen: Nit: Is it possible to move this into a separate script?  My general sense is that it's a bit lost in `package.json`.

gulp/config.js
```diff
@@ -63,11 +46,7 @@ module.exports = {
 
         // documentation site
         pldoc_src: pldocSrc + '/static/images',
-        pldoc_src_files: pldocSrc + '/static/images/**/*',
-
-        // example demo
-        demo_src: demoSrc + '/static/images',
-        demo_src_files: demoSrc + '/static/images/**/*'
+        pldoc_src_files: pldocSrc + '/static/images/**/*'
```
bjacobel: This is just a +diff because I removed the comma at the end of the line, but yep - we should absolutely be using camelCase for these (or removing them if they're no longer used, which I think these `fonts` ones might not be). I'll see what I can do with cleaning this file up :thumbsup:

package.json
bjacobel: `"scripts": {}` is a pretty common thing for a package to configure build/development tasks if it's not using another tool like gulp, so I guess the question is does having an `npm start` command that does the same thing as `$(npm bin)/gulp` add any value?

(I added this because I'm not a huge Gulp fan and I kind of just wanted to prove we didn't need it to do the development workflow... so having proved my point, I'll remove this and just make it an alias on my machine if nobody but me is going to use it 😛)

webpack.config.js
```diff
@@ -3,49 +3,99 @@
 'use strict';
 
 var path = require('path'),
-    Webpack = require('webpack'),
-    outputRoot = process.env.OUTPUT_ROOT !== undefined ? process.env.OUTPUT_ROOT : 'pldoc/public/',
-    siteRoot = process.env.SITE_ROOT !== undefined ? process.env.SITE_ROOT : '/',
-    publicJavaScriptRoot = 'public/';
+    webpack = require('webpack'),
+    ExtractTextPlugin = require('extract-text-webpack-plugin');
```
bjacobel: Added a bunch more documentation to this file about what all the different settings and paths are doing, let me know what you think.

gulp/config.js
bjacobel: Yep, this file ended up being used in like two places. I think it's adding more complexity than it's solving, so it's gone now!

package.json
bjacobel: @dsjen I removed this, as I decided providing two paths to the same functionality isn't good design.

dsjen:
@bjacobel -- thank you for responding to my feedback!  I'm running the page using webpack and it's looking great for the most part.  I did run into a problem where the page returns a `404` for `http://localhost:8080/public/pattern-library-doc.css` with the default gulp command.  The styles looked to be loading fine though.  To `generate pattern-library-doc.css`, I ran a production build to generate the assets for the `public` directory.  Any ideas?

bjacobel:
@dsjen, yep, I talk about the 404 for the css file in the "Known Issues" section of the PR. We can look at fixing it in the future, but it'd require creating a new Jekyll theme, I think.

dsjen:
@bjacobel -- I see!  Thanks for the clarification.  :)  I guess a jira ticket would be nice to capture that so that we can address it later.

LGTM!  👍

PR #369, opened 2016-10-26
Susy-less grids

bjacobel:
## Description

[TNL-5781](https://openedx.atlassian.net/browse/TNL-5781)

Develop an API-compatible replacement for Susy in the pattern library (based on Flexbox).

First draft. Some things are probably broken or done in ways that don't make sense.

http://ux-test.edx.org/bjacobel/grids/

EDIT: I'm also spinning up a sandbox to demo installing this into Platform. https://grids.sandbox.edx.org

Pages to look at:
- [Grid System](http://ux-test.edx.org/bjacobel/grids/design_elements/grid/)
- [Grid System Detailed Example](http://ux-test.edx.org/bjacobel/grids/examples/grid/)
- [Layouts](http://ux-test.edx.org/bjacobel/grids/design_elements/layouts/)
- [Layouts Detailed Example](http://ux-test.edx.org/bjacobel/grids/examples/layouts/)

Things have been touched throughout the pldoc site (as it uses the pattern lib for its own layout, and was probably the app most dependent on the old API out of all of the places we use the Pattern Lib...) but those four pages use most things.

## Testing Checklist
- [ ] Manually test right-to-left behavior.
- [ ] Manually test a11y support.

## Non-testing Checklist

## Post-review

## Reviewers
- [x] @alisan617 
- [x] @andy-armstrong 
- [x] @dsjen 
- [ ] @thallada 
- [x] @AlasdairSwan


bjacobel:
For those interested in the performance of this: 

```
 wheatley > ~ > code > edx-pattern-library > bjacobel/grids > $ > atime 10 $(npm bin)/node-sass --include-path=./node_modules --output=/tmp/out.css pldoc/static/sass/pattern-library-doc.scss
averaging 10 runs:
real: .693 sec
user: .583 sec
sys: .084 sec
```

My next step is getting this to work in Platform and taking some benchmarks there.

pattern-library/sass/patterns/_grid.scss
```diff
@@ -14,276 +14,180 @@
...
-// LAYOUT: set debug settings
-$susy-debug: layout($susy-default show);
-
-
-// ----------------------------
-// #GLOBALS
-// ----------------------------
-// set basic grid settings for app
-@include layout($susy-default);
+@mixin span($cols, $margin:'') {
```
alisan617: Could the name 'span' be more descriptive?

bjacobel: Usage of `span` comes from Susy; their `span` function is actually what was making compilation super slow, I think. I decided to keep the name (and keep it mostly API-compatible) so that migration to this new version will be easier.

e.g., there's about 30-40 places in edx-platform using `span` right now, many of them will just keep working with this new version of `span` (some of them will need changes to the arguments passed).

<details>
```
wheatley > ~ > code > edx-platform > master > $ > git grep "span(" *.scss
cms/static/sass/elements-v2/_header.scss:    @include span(12);
cms/static/sass/elements-v2/_header.scss:    @include span(7);
cms/static/sass/elements-v2/_header.scss:    @include span(4 last);
cms/static/sass/elements-v2/_sock.scss:    @include span(12);
cms/static/sass/elements-v2/_sock.scss:      @include span(8);
cms/static/sass/elements-v2/_sock.scss:      @include span(4);
common/static/sass/bourbon/css3/_columns.scss:@mixin column-span($arg: none) {
common/static/sass/neat/functions/_private.scss:@function container-span($span: $span) {
common/static/sass/neat/grid/_span-columns.scss:  $container-columns: container-span($span);
lms/static/certificates/sass/_components.scss:        width: span(10);
lms/static/certificates/sass/_layouts.scss:        @include span(6 first);
lms/static/certificates/sass/_layouts.scss:        @include span(6 last);
lms/static/certificates/sass/_layouts.scss:            width: span(12);
lms/static/certificates/sass/_layouts.scss:                width: span(10);
lms/static/certificates/sass/_layouts.scss:           width: span(12);
lms/static/certificates/sass/_layouts.scss:               width: span(8);
lms/static/certificates/sass/_layouts.scss:            width: span(6);
lms/static/certificates/sass/_layouts.scss:                width: span(3);
lms/static/certificates/sass/_layouts.scss:                width: span(2);
lms/static/certificates/sass/_layouts.scss:                width: span(2);
lms/static/certificates/sass/_layouts.scss:            width: span(12);
lms/static/certificates/sass/_layouts.scss:                width: span(4);
lms/static/certificates/sass/_layouts.scss:                width: span(3);
lms/static/certificates/sass/_layouts.scss:                width: span(3);
lms/static/certificates/sass/_print.scss:            width: span(8);
lms/static/certificates/sass/_print.scss:            @include span(12 last);
lms/static/certificates/sass/_views.scss:      @include span(9 of 12 first);
lms/static/certificates/sass/_views.scss:      @include span(3 of 12 last);
lms/static/sass/elements/_program-card.scss:  @include span(12);
lms/static/sass/elements/_program-card.scss:    @include span(6);
lms/static/sass/elements/_program-card.scss:    @include span(6);
lms/static/sass/elements/_program-card.scss:    @include span(6);
lms/static/sass/shared-v2/_base.scss:    width: span(12);
lms/static/sass/shared-v2/_footer.scss:            @include span(12);
lms/static/sass/shared-v2/_footer.scss:                @include span(8);
lms/static/sass/shared-v2/_footer.scss:                    @include span(9);
lms/static/sass/shared-v2/_footer.scss:            @include span(4);
lms/static/sass/shared-v2/_footer.scss:            @include span(12);
lms/static/sass/shared-v2/_footer.scss:                @include span(4);
lms/static/sass/shared-v2/_modal.scss:    @include span(5);
lms/static/sass/views/_program-list.scss:  @include span(12);
lms/static/sass/views/_program-list.scss:    @include span(9);
lms/static/sass/views/_program-list.scss:  @include span(12);
lms/static/sass/views/_program-list.scss:    @include span(3);
</details>

**Comment by talbs** [comment]
I don't really have a horse in this race anymore, but FWIW http://flexboxgrid.com/ looks solid if you're looking for a more simple implementation of a grid system. If you're going for a more complex one, I'd recommend Tachyons' grid system (http://tachyons.io/docs/layout/grid/). Its modular, can be installed on its own, and gives you responsive utilities.

Good luck, gang!

File: pattern-library/sass/global/_settings.scss [hunk 2]
```diff
@@ -33,11 +33,8 @@ $em-base: 16; // deliberately sets bourbon-based em-base (http://bourbon.io/docs
 // grid - dimensions
 $grid-max-width:                rem(1170) !default;
 $grid-column-count:             12 !default;
-$grid-gutter-width:             (1/3) !default;  // set to fraction to be proportional to fluid column widths (see - http://susydocs.oddbird.net/en/latest/settings/#gutters)
-$grid-container-position:       center !default;
-$grid-gutter-position:          after !default;
-$grid-manual-gutter-position:   inside !default;
-$grid-column-width:             false !default;  // set to false for fluid columns by default (see - http://susydocs.oddbird.net/en/latest/settings/#column-width)
+$grid-gutter-width:             rem($em-base) !default;
+$grid-direction:                row !default; // set to 'row-reverse' to reverse. Follows flexbox spec.
```
andy-armstrong: Does this work with right-to-left?

pattern-library/sass/patterns/_grid.scss
```diff
@@ -14,276 +14,180 @@
...
+    align-items: flex-start;
+    box-sizing: border-box;
+    margin: 0 auto;
 }
 
-%grid-gallery {
-    @include break();
-    @include clearfix();
+@mixin grid-row {
+    @include span(12, none);
```
andy-armstrong: Did you want to make 12 a variable? There was some discussion of wanting to support 10 columns too.

pattern-library/sass/patterns/_layouts.scss
```diff
@@ -3,97 +3,29 @@
...
-// CASE: View-based layout with reversed columns
-%layout-view-reversed {
-    @include container($susy-view-reversed);
-}
-
 // basic layout column
-%layout-column {
-    @include span(full); // plan for mobile first
+@mixin layout-column {
+    @include span(12); // plan for mobile first
```
andy-armstrong: Should this be a variable?

pattern-library/sass/patterns/_layouts.scss
```diff
@@ -3,97 +3,29 @@
...
 .layout-full {
-    @extend %layout-default;
+    @include grid-container;
 
     .layout-col-a,
     .layout-col-b,
     .layout-col-c,
     .layout-col-d {
-        @include span(full);
+        @include span(12);
```
andy-armstrong: Should this be a variable?

pattern-library/sass/patterns/_layouts.scss
```diff
@@ -212,216 +138,203 @@ $susy-view-reversed:    layout($susy-default auto $grid-direction-reversed); //
...
+        @include grid-container;
+        @media(min-width: $bp-screen-md) {
             .layout-col-a {
-                @include span(3 of 12 first);
+                @include span(3, before);
             }
 
             .layout-col-b {
-                @include span(6 of 12);
+                @include span(6, none);
```
andy-armstrong: I'm not so sure that my previous review comment of making 12 a variable makes sense, in that you'd then to do math in all these situations. Maybe it is simpler to be hard-coded for now.

andy-armstrong:
I don't think this is your change, but the buttons on the examples do not use the pattern library. It seems that the class ``btn-primary`` is no longer valid (I think Chris changed this):

http://ux-test.edx.org/bjacobel/grids/examples/grid/

pattern-library/sass/patterns/_grid.scss
bjacobel: Yep, this one needs to be a variable - it should fill the whole row regardless of what you have  specified as your default number of grid rows, so I'll use that variable here.

pattern-library/sass/patterns/_layouts.scss
bjacobel: IMO this one should not be a variable, because it's a demo of one possible layout you can create.

pattern-library/sass/patterns/_layouts.scss
bjacobel: Same as above - these classes won't be consumed by others, they're just a demo here on this page, so I don't think a variable is needed.

pattern-library/sass/patterns/_layouts.scss
andy-armstrong: I didn't realize that there are layouts that are only provided as demos. It would be good to move demos out into pldoc, and only leave documented patterns here.

bjacobel: Turns out this isn't actually used anywhere, so I'm deleting it.

pattern-library/sass/patterns/_layouts.scss
bjacobel: So I thought about this a bit, and I think anywhere where we're providing classes you can consume directly we should stick with 12 as the basis (and thus hardcode fractions of 12) because it's assumed that if you're using those classes, you don't want to do any customization.

However, in places where we're providing mixins or extendable classes for you to use to build your own layout system, we should use variables, because we want to support overriding the default number of columns for those cases.

bjacobel:
@andy-armstrong It's `btn-base` that's been removed and was causing funkiness on those buttons, I've fixed that issue in 2244b55.

Preview site updated with the changes I've made so far.

bjacobel:
@dsjen @thallada Do you think Insights will be able to install this and try it out this week? Would like to just confirm that the grid classes are basically working before merging this.

pattern-library/sass/patterns/_grid.scss
dsjen: I think a description of what the mixin is doing would be helpful.

pattern-library/sass/patterns/_grid.scss
```diff
@@ -14,276 +14,176 @@
...
+        width: calc(#{cols($cols)} - 0.5 * #{$grid-gutter-width});
+    } @else if ($margin == 'none') {
+        @include margin-left(0);
+        @include margin-right(0);
+        width: cols($cols);
+    }
+}
 
+@mixin pre($cols) {
+    @include margin-left(cols($cols) !important);
```
dsjen: `!important` make me a bit nervous.  Is it necessary?

bjacobel: That's a very valid anxiety, I throw up red flags when I see `!important`s too.

The `pre` and `post` mixins are used in grid columns to give them a certain amount of empty columns before or after that column. The best example of how they're used is these guys:



Each of those boxes looks in Sass like this:

```sass
.col-H {
    @include span(4, none);
    @include pre(8);
}
```

which when it's compiled out to CSS looks something like this: 

```css
.col-H {
    margin-left: 0;
    margin-right: 0;
    width: 33.333%;
    margin-left: 66.666% !important;
}
```

So, as you can see, in that case, the `!important` isn't necessary - the `pre`'s `margin-left: 66%` would override the one set on the same selector at the same specificity just by virtue of coming after it in the stylesheet. But if you messed up and wrote:

```sass
    @include pre(8);
    @include span(4, none);
```

Then the `margin-left` from your `pre` would be overridden by the `margin-left: 0` from the `span` mixin, and `pre` wouldn't work at all. So I added the `!important` because 1) I wanted these mixins to be able to be used on the same selector (which maintains compatibility with the existing Susy API) and 2) so that I wouldn't create a situation where the order in which you used the mixins could have an effect.

I think the only way around this would be to use a pseudoelement `:before` with `pre()` (and `:after` with `post()`), rather than margins, but I played around with that for a little bit and couldn't figure out how to make pseudoelements work well with flexbox.

bjacobel:
@dsjen How did Insights look when you updated to this version?

dsjen:
@bjacobel --testing right now.  I'll report back in a bit!

dsjen:
I believe https://github.com/edx/ux-pattern-library/blob/master/bower.json#L58 needs to be removed too.

bjacobel:
Updated PR with the change that underfilled grids stack to the left rather than distributing empty space between. I deleted some of the "perfect" rows in the pldoc Grid System page (updated on ux-test.edx.org) so this behavior can be seen.

andy-armstrong:
👍  for the changes since I last approved this!

bjacobel:
@alisan617 @dsjen @thallada @AlasdairSwan Would be great to get a final review and a second thumb from one of you today! I've already had a few people ask me when this is landing in Platform after I demoed it on Friday :)

alisan617:
I took a quick look in the sandbox. Things look good. 👍

AlasdairSwan:
Looks good @bjacobel :shipit:

dsjen:
I'm looking now.  Thanks for pinging!

PR #371, opened 2016-11-10
Fix issue with prebuilt layouts below 768px

bjacobel:
## Description

Prebuilt layouts at screen sizes < 768 px were broken after my most recent UXPL PR (new grid system).

It's pretty obvious, don't know why I didn't catch it. Live on http://ux.edx.org/design_elements/layouts/

Preview sandbox with fix: http://ux-test.edx.org/layouts-responsive-fix/design_elements/layouts/

This will need to go out as 0.17.1 and be integrated into platform before the next release as it's causing [TNL-5934](https://openedx.atlassian.net/browse/TNL-5934)

## Reviewers
- [x] @alisan617 
- [ ] @andy-armstrong

### DIFF
--- a/pattern-library/sass/patterns/_layouts.scss
+++ b/pattern-library/sass/patterns/_layouts.scss
@@ -24,6 +24,10 @@
     }
 }
 
+.layout-col {
+    @include grid-row;
+}
+
 // 1/4 + 3/4 Layout
 .layout-1q3q {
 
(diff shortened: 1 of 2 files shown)


alisan617:
👍

bjacobel:
@andy-armstrong Can you review this today? Thanks.

andy-armstrong:
@bjacobel I don't understand what the expected behavior is. It seems to me that the pre-built layouts stop being full width at the smaller breakpoint, which doesn't seem correct.

Here's how it looks before the break:

image

Then as it hits the break, all the boxes become too small:

image

I would expect them to look just like the screen shot before the break.

andy-armstrong:
@bjacobel also, the colors seem to change. I wonder if this is a pldoc styling issue rather than one with the Pattern Library.

